PR #4, opened 2019-09-15
Added support for SVG attributes

macprog-guy:
Hi Ryan,

I have added the list of all SVG attributes that don't contain dashes and have added them to the list of properties in your index file. I've also added a test for a rect element with both static and computed values. It does not seem to break any existing tests in the babel plugin either.
Cheers,

### DIFF
--- a/index.js
+++ b/index.js
@@ -91,9 +91,6 @@ Attributes = {
   calcMode: {
     type: Types.ATTRIBUTE
   },
-  class: {
-    type: Types.ATTRIBUTE
-  },
   clip: {
     type: Types.ATTRIBUTE
   },
@@ -209,9 +206,6 @@ Attributes = {
     type: Types.ATTRIBUTE
   },
   height: {
-    type: Types.ATTRIBUTE
-  },
-  href: {
     type: Types.ATTRIBUTE
   },
   hreflang: {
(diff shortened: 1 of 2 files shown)


ryansolid:
Awesome, I will take a closer look this evening. I was debating between 2 approaches. The one you did here, and the other detecting which elements were SVG by tag and just defaulting to attributes generally. The latter is the approach Surplus uses. I think what it's going to come down to me is which has the least potential impact on Custom Elements (Web Components). I think it might be fair to just detect Custom Elements as well.

Basically I see there is a limit to this global attribute thing I've started. If someone happens to name a custom element property the same as SVG property unexpected behavior etc..

ryansolid:
Yeah, being able to resolve more compile-time without referencing every attribute seems the best (svg element list). So going to spend more time with this, but it is now near the front of the list. I'm doing some big changes in the babel plugin right now. I expect a solution in the next minor version (0.12.0). But it will probably be a week or so.

ryansolid:
I ended up going with the approach mentioned in my last comment so I won't be merging this. Thanks so much for the PR. Basic SVG support is available in Solid v0.10.10 with Babel JSX DOM Expressions 0.11.6.

macprog-guy:
Hi Ryan, thanks for adding the SVG support in. I'll be using it pretty soon ;-)

PR #9, opened 2020-03-07
optimizations: group JSX options, conditionally assign JSX options vi…

yuens1002:
new to solid, my attempt to optimize existing code to get a better feel for the current code base.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/index.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/index.js
@@ -5,22 +5,31 @@ import VoidElements from "./VoidElements";
 
 export default babel => {
   const { types: t } = babel;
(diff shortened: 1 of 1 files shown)


yuens1002:
not sure what the issues are the expectations are mathing except in the import statements where "r-" is added.

yuens1002:
ran npm run test locally against untouched master, tests fails for this same file.  perhaps something is going on with the test cases?

ryansolid:
It's possible.. I've been upending stuff.. I think the babel plugin is getting cached. I will check it out tonight.

ryansolid:
Yeah it was just two small typos that lead to the vars never being assigned. So while it didn't crash the vars were always the default. Fixed now.

PR #14, opened 2020-05-29
Fix ref with arrow functions and other non-lvals

ricochet1k:


ryansolid:
Thank you. No idea where the test went for this as I had `forwardRef` previously with the same functionality.

PR #16, opened 2020-06-18
Allow props to be null or undefined, for parcel.js

hugojosefson:
Using parcel.js with the following jsx pragma and code:

```jsx
/** @jsx h */
import h from 'solid-js/h'
import { render } from 'solid-js/dom'

render(
  () => <h1>HELLO</h1>,
  document.getElementById('root')
);
```
...causes it to call `h` like this:

```js
h("h1", null, "HELLO")
```

...which fails, because `props` is `null`, and is seen as being `typeof "object"`.

This commit allows `props` to be ignored if `null`, instead of being treated like an `object`.

### DIFF
--- a/packages/hyper-dom-expressions/src/index.ts
+++ b/packages/hyper-dom-expressions/src/index.ts
@@ -62,7 +62,7 @@ export function createHyperScript(r: Runtime): HyperScript {
           let props: Props = {},
             dynamic = [],
             next = args[0];
-          if (typeof next === "object" && !Array.isArray(next) && !(next instanceof Element))
+          if (typeof next === "object" && next != null && !Array.isArray(next) && !(next instanceof Element))
             props = args.shift();
           for (const k in props) {
             if (typeof props[k] === "function") dynamic.push(k);


ryansolid:
Thanks although as far as I can tell this is only a problem with function components and the "h1" in the example above should work fine. The fix also only addresses function components so I want to make sure I'm not missing another issue here.

And for my own curiousity, why not use the significantly more optimized babel plugin instead of HyperScript? If it's a matter of TypeScript support with Parcel, Parcel 2 alpha the next branch supports TypeScript with Babel transformations.

hugojosefson:
Thank you @ryansolid for your quick response, and release!

You are absolutely correct; this issue was actually with function components, and it did indeed not appear when only using `h1` as in my initial example above. I confused myself when typing it up, half based on code I was writing with function components :D

Just to verify, I set up an actual (minimal) example:
https://github.com/hugojosefson/hyper-dom-expressions-16

It fails on the previous `hyper-dom-expressions@=0.18.11` as we expected, and works correctly with your new release of `hyper-dom-expressions@=0.19.0`!

hugojosefson:
The reason for me to not use babel and for example [solid-scripts](https://github.com/ryansolid/solid-scripts) via [create-solid](https://github.com/ryansolid/create-solid), is not related to TypeScript. Almost the opposite!

I am interested in seeing how good of a developer experience I can figure out with not using any build tools, running the server as pure esmodules in Node.js 14+, and using pure esmodules in the browser as well. No bundling or packaging. So that's why I look into simpler helpers such as `hyper-dom-expressions`.

My goal is to find a sweetspot with only using a few libraries, that don't clutter the understanding we can have of the code that we write. This is in contrast to how complex React and Redux can become when we don't fully understand what the code that we see in the file in front of us does, as was mine and a colleague's experience at work in the previous weeks.

hugojosefson:
Using JSX with `hyper-dom-expressions` was perhaps a small step away from my completely minimalist challenge :) but I thought it would be interesting to see how much/little complexity would be necessary to bring JSX syntax into the setup I'm trying out in [hugojosefson/chemical-example#master](https://github.com/hugojosefson/chemical-example) and [#solid](https://github.com/hugojosefson/chemical-example/tree/solid).

ryansolid:
Yeah the default JSX stuff should work there. The big issue is that you have be very conscious of what you wrap in functions. Reactive things always run in context so it requires a lot higher understanding to get started. With the babel JSX plugin I wrote you just write it like React and not have to worry about it. That and the performance degradation is why I almost never recommend it.

Tagged Template Literal version is actually considerably more performant than the HyperScript version. So if you want that sort of experience it's probably better. Mind you it still have a few gaps like no spreads. But JSX compiler on HyperScript is sort of worst of both worlds because you don't get to benefit from the better DX experience or have the best performance, and you still have a build step.

PR #17, opened 2020-06-19
 Log value separately 

junaid1460:
## Description
Log value separately, console will handle it's visualisation

## Changes
- [ ] Bug fix
- [x] Improvement
- [ ] Feature


ryansolid:
Oh never even thought about that. I will keep that in mind in the future thank you.

PR #21, opened 2020-08-08
fix: transform callExpression ref

mduclehcm:
fix: https://github.com/ryansolid/solid/issues/203


mduclehcm:
I have a problem with window newline! I don't know how to fix this. =.=

ryansolid:
Awesome.. I had already fixed this in my branch before I even saw your intention to fix this in the other thread. I did something very similar but I like this solution even better. Thanks.

PR #22, opened 2020-08-12
update html template tag type signature

trusktr:


packages/lit-dom-expressions/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "declaration": true,
     "target": "esnext",
     "moduleResolution": "node",
-    "strict": true
+    "strict": true,
+    "lib": ["es2018"]
```
trusktr, line 9: The definition comes from es5, but it's reasonable to include type defs for ES from 2 years ago to be safe. That's what I've been doing, is just using whatever 1 or 2 year-old target.

Hmmm... But I see you're targetting `esnext` in the target field. That means basically compile nothing. Not sure that's a safe bet (unless other people have a build step).

ryansolid:
Thank you.

PR #26, opened 2020-08-22
move dom-expressions to package.json "dependencies"

### LINKED ISSUE #25: [lit-dom-expressions] dom-expressions should be a dependency of lit-dom-expressions.
Otherwise TypeScript users will get errors like:

```
node_modules/lit-dom-expressions/types/index.d.ts:1:83 - error TS2307: Cannot find module 'dom-expressions/src/runtime'.

1 import { effect, insert, createComponent, delegateEvents, classList, style } from "dom-expressions/src/runtime";
```

trusktr:
fixes #25

### DIFF
--- a/packages/lit-dom-expressions/package.json
+++ b/packages/lit-dom-expressions/package.json
@@ -18,8 +18,10 @@
     "test:coverage": "npm run build && jest --coverage",
     "prepublishOnly": "npm run build"
   },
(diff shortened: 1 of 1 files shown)


ryansolid:
I'm just going to remove the need for the dep. I've done that in the latest. So the PR no longer is necessary. Thanks.

PR #28, opened 2020-09-10
Remove g flag from escape regex, fixes #27

frenzzy:
See https://github.com/ryansolid/dom-expressions/pull/27#issuecomment-690537305 for more info


ryansolid:
Thank you. In theory this is a less greedy search so it should be even more performant. I look forward to testing it.

PR #31, opened 2020-09-26
add hypercache as a new renderer

modderme123:


ryansolid:
Decided to drop it? I wasn't clear if you going to change the mechanism or not. In any case it was an interesting experiment. I think we learned some good things from this.

modderme123:
I agree that I learned a lot making this. I closed it because the ergonomics weren't better than the tagged template and neither was the performance. If I can get one to be better I'll reopen or make a new pr.

PR #36, opened 2020-10-14
Typings for generateHydrationScript({ resolved })

frenzzy:
Just found a [usage](https://github.com/ryansolid/solid/blob/e99f2ad590cad09a9ab1983369f097b6c16bf666/packages/solid-ssr/examples/async/index.js#L29) but not the typings:
https://github.com/ryansolid/dom-expressions/blob/ba1ea6b2d8136d9fcae412682ec162bc010f6c50/packages/dom-expressions/src/runtime.js#L329-L333


ryansolid:
Thank you

PR #39, opened 2020-11-07
Class list: accept array

MrFoxPro:


MrFoxPro:
Take a look at https://github.com/jorgebucaran/classcat/blob/master/src/index.js

ryansolid:
There are few considerations here so might take me some time to get through this.

1. If we aren't replacing the full string we need to do a negative diff too against the prev value if available, this will probably mean this change doubles the code in the helper.
2. I was planning to look at statically analyzing syntax here, so I need to take that into consideration.
3. I was wondering if directives change the game here. I actually don't like that there is a special helper here. Needs consideration everywhere. There is the option to just make this `class` handling. Alternatively may the object syntax should just be `class:some-class={state.visible}` and no special object case. Or maybe this whole thing belong in a `use:classList` custom directive.

In all cases need to weigh those with the usefulness of the array syntax over `${"a"} ${"b"} ${"c"}` which we get for free. Although admittedly it probably handles the falsy case better.

PR #54, opened 2021-02-26
Partially revert 86c9ff1 Add static template hoisting to SSR

nicketson:
In 86c9ff1, state for templates and events was moved from `path.scope.data` to `path.state`. Newer versions of babel have a bug with the way that `TraversalContext`s are tracked and it can cause `path.state` to be clobbered. This issue manifests itself when other babel plugins (notably `@babel/plugin-proposal-class-properties`) are used that visit the same AST nodes. This pull request moves template and events state back to `path.scope.data`.

The issue seems to have first occurred after this https://github.com/babel/babel/pull/12302
I believe this unfinished PR will fix the `TraversalContext` bug https://github.com/babel/babel/pull/12634

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -342,7 +342,9 @@ function transformAttributes(path, results) {
           ) {
             // can only hydrate delegated events
             hasHydratableEvent = true;
-            const events = path.state.events;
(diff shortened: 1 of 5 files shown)


ryansolid:
Ah.. I'd seen some weird stuff recently but didn't know where it was coming from. We make heavy use of state in Marko so I was really liking the convenience of the approach but I can revert for now I guess.

PR #55, opened 2021-03-02
Add setAttributesWrapper and dynamicExpressionWrapper config options

nicketson:
This generalizes babel-plugin-jsx-dom-expressions generated calls to `effect()` and `memo()`. The function names are now configurable with new options `setAttributesWrapper` (effect) and `dynamicExpressionWrapper` (memo). The plugin behavior is unchanged when the default options are used. Additionally, either option can be set to `false`, which will essentially treat expressions as if they were static and won't wrap them. 

The use case for setting the new options to `false` is to use babel-plugin-jsx-dom-expressions with a wrapperless non-reactive runtime.

I've copied a subset of relevant tests from `dom.spec.js` to `dom-wrapperless.spec.js` with the new options set to false.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/config.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/config.js
@@ -8,5 +8,7 @@ export default {
   requireImportSource: false,
   wrapConditionals: true,
   contextToCustomElements: false,
-  staticMarker: "@once"
+  staticMarker: "@once",
+  setAttributesWrapper: "effect",
+  dynamicExpressionWrapper: "memo",
 };
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -437,7 +437,7 @@ function transformAttributes(path, results) {
             }
           }
         } else if (
-          isDynamic(attribute.get("value").get("expression"), {
+          config.setAttributesWrapper && isDynamic(attribute.get("value").get("expression"), {
             checkMember: true
           })
         ) {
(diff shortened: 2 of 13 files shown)


ryansolid:
Hmm.. does this mean you use a different runtime? Since the runtime depends on those primitives, like `insert` uses `effect` etc. I mean that's all fine I just want to make sure I understand the scenario so that by supporting this we won't be tracking down a number of other downstream things, and to keep in mind the implications of this option for future development I do.

nicketson:
> Hmm.. does this mean you use a different runtime? Since the runtime depends on those primitives, like `insert` uses `effect` etc. I mean that's all fine I just want to make sure I understand the scenario so that by supporting this we won't be tracking down a number of other downstream things, and to keep in mind the implications of this option for future development I do.

Yup. I have a (in progress) fork of dom-expressions. Basically just a subset of `client.js` without calls to `effect`, reconcileArrays, etc.

ryansolid:
I've been thinking of how I'd want to best handle this. But I think this is close enough. I might tweak the naming but I will introduce the mechanism. Thanks for the PR.

nicketson:
> I've been thinking of how I'd want to best handle this. But I think this is close enough. I might tweak the naming but I will introduce the mechanism. Thanks for the PR.

No problem. Thanks for merging! And yes, I agree the naming could be better.

ryansolid:
It's mostly that you named it literally to what the compiler is doing but it does other stuff in the runtime, and since I don't imagine your use case is as common I'd rather it reflect the intention of the primitive rather than the application in the compiler. In any case I should have published to npm so you can give it a try. I went with effectWrapper and memoWrapper.

PR #57, opened 2021-03-19
Fix inaccurate docs

nicketson:
The `createComponent()` examples in the documentation have an unused argument `dynamicKeys`. `babel-plugin-jsx-dom-expressions` doesn't generate any calls to createComponent with this argument. I'm assuming this is from an old version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ function memo(fn, equal) {
   return s;
 }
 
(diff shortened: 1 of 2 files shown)


ryansolid:
You are correct. I built that into the compiler in a later version. Thank you.

PR #73, opened 2021-06-12
Make TS user friendly with custom tags and attributes

amoutonbrady:
This PR aims to make typescript way less in the way.

Instead of having to maintain a bazillion of different use case, this allows us to use what already exist and default to `any` for the rest (custom attributes, tags, etc..)

Before merging it could be interesting to have @trusktr opinion, but from my POV, this would allow us to have a break with TS when 1.0 launch.

---

The origin of this is the recent release of windicss 3.0, which added custom attributes to style your markup https://windicss.org/posts/attributify.html. I expect people bumping into using `tsx` files and wondering why TS is all over them because they use custom attributes such as `div text="green-500`.

This also allows for custom element to at least not yield an error. It won't be typed but at least they will be usable.

### DIFF
--- a/packages/dom-expressions/src/jsx.d.ts
+++ b/packages/dom-expressions/src/jsx.d.ts
@@ -75,7 +75,7 @@ export namespace JSX {
   type OnCaptureAttributes<T> = {
     [Key in keyof CustomEvents as `oncapture:${Key}`]?: EventHandler<T, CustomEvents[Key]>;
   }
-  interface DOMAttributes<T> extends CustomAttributes<T>, ActionAttributes, PropAttributes, AttrAttributes, OnAttributes<T>, OnCaptureAttributes<T>, Record<string, any> {
+  interface DOMAttributes<T> extends CustomAttributes<T>, DirectiveAttributes, PropAttributes, AttrAttributes, OnAttributes<T>, OnCaptureAttributes<T>, Record<string, any> {
     children?: Element;
     innerHTML?: string;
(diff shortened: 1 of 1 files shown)


aminya (changes requested):
Instead of extending `Record<string, any>`, it is common to add a property that opens up the interface for more properties.
```ts
  [customProperty: string]: string // something specific like string not "any"
```

`Recrod<string, any>` means that they can pass "any" value to the functions, which I don't think is a correct approach.

ryansolid:
I don't know if this is a good thing. It means that any misspelled attribute wouldn't warn you. I think most TS users would think this is undesirable. I may not be a big fan of TypeScript but I think extending interfaces probably makes more sense. I realize this is tedious, but so is TypeScript. That's why you use it.

amoutonbrady:
Yeah that's totally understandable, I suppose this is also possible in user land to create a `.d.ts` file and extends the necessary interfaces.

I do however think that if we want to go down that path, we might need to document it. A typescript section in the doc with common typescript tricks and tips could be useful. Anyway, that's a discussion for another issue.

ryansolid:
Probably. Although that is the same for every JSX based language. But people will hit it more with Solid (or Stencil) given how TS heavy we are.

ryansolid:
Yeah I'm pretty confident this not the path forward. TS users seem to prefer strictness and explicit definitions.

amoutonbrady:
Yeah that makes sense

PR #75, opened 2021-07-09
Fix events firing on disabled elements

pavulon:
Makes it consistent with vanilla (without event delegation) and React

https://codesandbox.io/s/xenodochial-resonance-mz4hd

### DIFF
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -297,7 +297,7 @@ function eventHandler(e) {
 
   while (node !== null) {
     const handler = node[key];
-    if (handler) {
+    if (handler && !node.disabled) {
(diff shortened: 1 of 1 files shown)


ryansolid:
Does disabled matter up the tree or only on the initial target? This would stop where ever disabled is applied. Is that consistent?

pavulon:
I believe it matters only up the tree. Disabled elements themselves don't produce events, but their descendants do, and they will bubble up without calling the handler on the disabled element. That what's this change is supposed to do, unless I've missed something?

ryansolid:
Yeah that sounds right. The disabled event won't fire the event so that's covered and it only applies to mouse events but I only delegate mouse events so I think we are good. Thanks.

PR #79, opened 2021-07-18
add missing "meter" role

atk:
The `role="meter"` is woefully underused, for example for amplitude visualization or password strength - and even missing here. More under: https://w3c.github.io/aria-practices/examples/meter/meter.html


ryansolid:
Thank you, I grabbed the list from MDN docs in the past. It seems like this is always a moving target so continue to put these forward should you ever find anything else of this nature.

PR #80, opened 2021-07-20
Handle white-space-only text correctly, fixes solidjs issue #571

atk:
If a text consists only of white-space characters (worst case: `"\n"`), the return value may be an empty string, which may cause issues like in SolidJS #571 and unneccessary filtering of non-data. Detecting this case and returning a simple space character fixes both.


ryansolid:
I actually already made a fix of just keeping hoisted literals.. well literal and not applying trim to them. https://github.com/ryansolid/dom-expressions/commit/0f01ded18826fbc3f06d07eafaa72de35b438cfa

I've considered just adding the space before but there are places I definitely want it. Thanks for putting this together.

PR #84, opened 2021-08-23
Fix typings for <details>

atk:
Typing details as DetailsHTMLAttributes<HTMLElement> will stop the open property from working correctly in typescript.


packages/dom-expressions/src/jsx.d.ts
```diff
@@ -19,7 +19,7 @@ export namespace JSX {
   interface ArrayElement extends Array<Element> {}
   interface FunctionElement {
     (): Element;
-  }
```
ryansolid, line 22: Was this removal intentional?

atk: No, it wasn't.

ryansolid:
Thank you

PR #90, opened 2021-09-21
Fix `ref` property to normalize `TsNonNullExpression` and `TsAsExpression`

LXSMNSYC:
This PR fixes a known issue for `dom-expressions` where using non-null and type-as expressions causes `ref` property to get stripped out of the output code.

TODO
- Should I add fixtures/snapshots for the TS expressions?

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -335,7 +335,6 @@ function transformAttributes(path, results) {
       ) {
         if (key === "ref") {
           // Normalize expressions for non-null and type-as
-          console.log(value.expression);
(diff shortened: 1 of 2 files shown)


ryansolid:
Hmm... That's interesting. I didn't expect TS to affect the compilation so I never set them up that way. Presumably, this doesn't break if typescript isn't present in the consuming project as the types are part of Babel Types. But getting the JSX types into the project is a bit awkward since DOM Expressions doesn't expose the types in the way that can be used with the jsxImportSource.

I don't know. I'm ok with this one as a one shot I think compared to trying to change the test setup at the moment.

LXSMNSYC:
I made the new fixtures work however it requires overwriting the old fixtures for TSX to work. Since the output now works, I don't think I have to add the new fixtures?

ryansolid:
Thanks

PR #99, opened 2021-12-15
Hotfix: `Identifier` linking

LXSMNSYC:
This PR fixes `Identifier` handling through out the babel plugin. The reason we want to do this is so that source maps can properly link to the original JSX source, however Babel does not allow us to move `JSXIdentifier` in places where `Identifier` is valid. I used [`babel-plugin-transform-react-jsx`](https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts) as a reference point. The main goal of this PR is to not fallback to `t.identifier` as much as possible for places where it can be preserved, this allows Babel to know where it moved.

TODO:
- [x] component identifier handling
- [x] `use:*` bindings
- [x] component attributes
- [x] auto-import handling

Notes:
- This PR might not involve template/dom handling as string literals are generally not possible to link.

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -17,6 +17,9 @@ jobs:
     with:
       node-version: 14
 
+  - name: Bootstrap packages
+    run: npm run bootstrap
+
   - name: Installing deps
     run: npm install
 
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "test:coverage": "lerna run test:coverage --parallel",
     "build": "lerna run build --concurrency=1 --stream",
     "publish:release": "lerna run build && lerna publish",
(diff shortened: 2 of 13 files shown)


atk:
Not handling the templates is perfectly valid; since they are still evaluated by the rendering, this should suffice to measure code coverage.

LXSMNSYC:
Okay so I think this is partially good enough. There are some parts of the code that I do not want to touch. I don't even know if there's a trick to map identifier to string literals.

ryansolid:
I merged this into Next. It looks good. Minor changes really and stoked to have better sourcemaps.

LXSMNSYC:
Thanks! I forgot to target `next` branch

PR #101, opened 2022-01-02
fix: add quotes around attribute names which are reserved words in js

fictitious:
Not sure if it was intended or not, but [this commit](https://github.com/ryansolid/dom-expressions/pull/99/commits/2b0b44d05ba2847c57fab25d1fa5ae21e21b1922) removed quotes in compiled JSX output for attribute names which are reserved words in javascript, like "class".

`get [class]()` without quotes is invalid syntax in chrome, so I think the quotes are necessary at least for the web. 

If it's ok to have quotes for all targets (ssr, universal?), the fix is pretty simple - revert `t.isValidIdentifier` call to the form it was before that commit - without the second `false` parameter.


ryansolid:
Thank you.

PR #102, opened 2022-01-05
Add href prop to <textPath>

0x241F31:
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/textPath


MrFoxPro:
Should I add this prop to all jsx defenitions in this repo?

ryansolid:
This is the right place. Not quite sure I know what you mean by "all".

PR #113, opened 2022-02-25
DirectiveFunctions alternative for typing directives

edemaine:
Directives (`use:___`) normally have two arguments, first the element/component and second the value passed in as a JSX attribute. The existing `Directives` overloadable interface only allows checks the value / second argument.  This PR provides a new overloadable interface `DirectiveFunctions` checks both (when present), by specifying the entire function type.

Example usage:

```ts
function model(element: HTMLInputElement, value: Accessor<Signal<string>>) {
  const [field, setField] = value();
  createRenderEffect(() => (element.value = field()));
  element.addEventListener("input", (e) => setField(e.target.value));
}

function log(element: DOMElement) {
  console.log(element);
}

let num = 0;
function count() {
  num++;
}

function foo(comp: number, ...args: string[]) {
}

declare module "solid-js" {
  namespace JSX {
    interface DirectiveFunctions {
      model: typeof model;
      log: typeof log;
      count: typeof count;
      foo: typeof foo;
    }
  }
}

{/*good:*/}
<input use:model={createSignal('')}/>
{/*bad:*/}
<input use:model/>
<input use:model={7}/>
<div use:model={createSignal('')}/>
{/*good:*/}
<div use:count use:log/>
{/*bad:*/}
<div use:foo/>
```

Testing tips would be appreciated. My best success was `npm link`ing `dom-expressions` into itself (`npm link; npm link dom-expressions`), creating a `tsconfig.json` and `jsx-runtime.d.ts`. I eventually got roughly the above file (without SolidJS imports and replacing `module "solid-js"` with `module "dom-expressions"`) to work and succeed/fail as intended, although the error for `<div use:model={'hi'}/>` is a bit weird (it says `use:model` can only take value `undefined` in this case, corresponding to an absent optional attribute, instead of complaining about the element type).

(I also tried `npm link`ing this into `solid` and putting tests into `src/web/test`, but I got all sorts of errors when trying to run `tsc` there... )

### DIFF
--- a/packages/dom-expressions/src/jsx.d.ts
+++ b/packages/dom-expressions/src/jsx.d.ts
@@ -57,12 +57,25 @@ export namespace JSX {
     $ServerOnly?: boolean;
   }
   interface Directives {}
+  interface DirectiveFunctions {}
   interface ExplicitProperties {}
   interface ExplicitAttributes {}
   interface CustomEvents {}
   interface CustomCaptureEvents {}
   type DirectiveAttributes = {
     [Key in keyof Directives as `use:${Key}`]?: Directives[Key];
+  };
+  type DirectiveFunctionAttributes<T> = {
+    [Key in keyof DirectiveFunctions as `use:${Key}`]?:
+      // If function has no arguments, value is arbitrary
+      Parameters<DirectiveFunctions[Key]>['length'] extends 0 ? any :
+      // Otherwise (at least one argument), check its type matches element type
+      T extends Parameters<DirectiveFunctions[Key]>[0] ? (
+        // If function has only one argument, value is arbitrary
+        Parameters<DirectiveFunctions[Key]>['length'] extends 1 ? any :
+        // Otherwise (at least two arguments), value must match second argument
+        Parameters<DirectiveFunctions[Key]>[1]
+      ) : never;
   };
   type PropAttributes = {
     [Key in keyof ExplicitProperties as `prop:${Key}`]?: ExplicitProperties[Key];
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah I don't have a great process for testing. I usually test this stuff independent of Solid.. Just put stuff in dom-expression tests and then test again when I run from Solid repo. Of course, I don't end up testing TS until the Solid repo which has lead to occasional extra fix commits after the fact.

packages/dom-expressions/src/jsx.d.ts
```diff
@@ -57,13 +57,26 @@ export namespace JSX {
...
+      // If function has no arguments, value is arbitrary
+      Parameters<DirectiveFunctions[Key]>['length'] extends 0 ? any :
+      // Otherwise (at least one argument), check its type matches element type
+      T extends Parameters<DirectiveFunctions[Key]>[0] ? (
+        // If function has only one argument, value is arbitrary
+        Parameters<DirectiveFunctions[Key]>['length'] extends 1 ? any :
+        // Otherwise (at least two arguments), value must match second argument
+        Parameters<DirectiveFunctions[Key]>[1]
+      ) : never;
+  };
```
otonashixav, line 79: I think this isn't correct; `Parameters<DirectiveFunctions[Key]>[1]` will give `Accessor<Signal<string>>` for the example provided, instead of simply `Signal<string>` as it should. 

I suggest this definition instead, which uses `infer` directly instead of using `Parameters`:

```ts
    type DirectiveFunctionAttributes<T> = {
      [K in keyof DirectiveFunctions as `use:${K}`]?: DirectiveFunctions[K] extends (
        el: infer E, // E will be unknown if not provided
        accessor: infer A // A will be unknown if not provided
      ) => void
        ? T extends E
          ? A extends Accessor<infer V>
            ? V // if it's an Accessor, the type should be the type the Accessor wraps
            : true // if it isn't, assume the second parameter isn't provided
          : never
        : never;
    };
```

`true` corresponds to `use:directive` with no parameters provided.

Users not typing the directive function with `Accessor` will simply see `type is not assignable to 'true'` which isn't very informative, but there's no way to throw a custom error here other than with a string literal (so instead of going `true` immediately, do `unknown extends A ? true : "directive parameters must be wrapped in an Accessor (() => your type here)"`)

otonashixav (changes requested):

packages/dom-expressions/src/jsx.d.ts
```diff
@@ -57,13 +57,26 @@ export namespace JSX {
     $ServerOnly?: boolean;
   }
   interface Directives {}
+  interface DirectiveFunctions {}
```
otonashixav, line 60: This could potentially restrict the types of directive functions allowed via an index signature:
```ts
    interface DirectiveFunctions {
      [x: string]: (el: Element, accessor: Accessor<any>) => void;
    }
```

Then later, filter out the index signature using `[Key in keyof DirectiveFunctions as string extends Key ? never : `use:${Key}`]?: ...`, or some other method. 

However there's no precedent for checking user-defined types elsewhere, so I'm not sure if it should be done.

PR #115, opened 2022-03-08
Mark `template` as `#__PURE__` in generated code.

novacrazy:
When components are included in files but are unused, Terser of course removes those functions. However, the generated calls to `template` outside of that function are kept, leading to code like:
```js
template("<div></div>");
const _tmpl$2 = template('<button type="button"></button>');
```
Note the first template is not assigned to anything, but Terser preserves it just in case it has side-effects.

However, the only side-effect `template` has is creating an unparented DOM node, which is just wasteful if not removed.

Therefore, marking `template` with the leading comment `/*#__PURE__*/` allows Terser the freedom to remove calls to it if the results are unused.

Also, may want to squash this when merging.


ryansolid:
Thank you.

PR #119, opened 2022-03-16
feat: add targeted event types

otonashixav:
This mostly allows event handlers to be declared as `function handleInput(ev: JSX.InputEvent<HTMLInputElement>): void`, whereas before only `const handleInput: JSX.EventHandler<HTMLInputElement, InputEvent> = (ev) => ...` would be valid if you wanted to have `currentTarget` and `target` narrowed. 

Compatibility with existing code is preserved, with existing instances of `*Event` replaced with `DOM*Event` outside the namespace referring to the original type, like `DOMElement` before them.

- I'm not sure whether this is the right place for these types, nor how to test these changes. 
- These types should be usable in the same way as react's event types, though only for actual DOM events. I'm not really familiar with react, so maybe someone could weigh in here. 
- This could be cut down to just adding ~~`SyntheticEvent`~~ `TargetedEvent` (as `Event`) without the rest of the types, which would be sufficient for doing `function handleInput(ev: JSX.Event<HTMLInputElement, InputEvent>): void`. In this case none of the `*Event` types in the namespace would need to be renamed to refer to the actual DOM events.
- ~~Is "SyntheticEvent" the right name for these types?~~ Is `TargetedEvent` the right name? Would `EventWithTargets` be clearer?
- Possibly these changes are unneeded as you could simply use e.g. `InputEvent & { currentTarget: HTMLInputElement }` as needed. 

Changed `SyntheticEvent` to `TargetedEvent` so as to not imply the event is synthetic, when in fact it is not.

### DIFF
--- a/packages/dom-expressions/src/jsx.d.ts
+++ b/packages/dom-expressions/src/jsx.d.ts
@@ -5,6 +5,20 @@
  * https://github.com/infernojs/inferno/blob/master/packages/inferno/src/core/types.ts
  */
 type DOMElement = Element;
+type DOMEvent = Event;
+type DOMClipboardEvent = ClipboardEvent;
+type DOMCompositionEvent = CompositionEvent;
+type DOMFocusEvent = FocusEvent;
+type DOMInputEvent = InputEvent;
+type DOMKeyboardEvent = KeyboardEvent;
+type DOMPointerEvent = PointerEvent;
+type DOMMouseEvent = MouseEvent;
+type DOMDragEvent = DragEvent;
+type DOMTouchEvent = TouchEvent;
+type DOMUIEvent = UIEvent;
+type DOMWheelEvent = WheelEvent;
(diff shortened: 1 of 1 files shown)


trusktr:
Since Solid JSX maps to basically real DOM, can we just change it so we can write stuff like

```js
function onClick(event: MouseEvent) {...}
```

using the real DOM types?

trusktr:
Oh nvm, that already works. https://playground.solidjs.com/?hash=-1362460424&version=1.3.13

What does this solve exactly?

otonashixav:
`MouseEvent` doesn't narrow `currentTarget`. Currently if you wanted to access it in the event handler you'd have two options, either manually do `MouseEvent & { currentTarget: HTMLButtonElement }`, or use the existing `JSX.EventHandler` type which only types the function, but not an event. https://playground.solidjs.com/?hash=1468161564&version=1.3.13

PR #120, opened 2022-04-04
move ref call outside of the spread effect

fictitious:
This fixes solid issue solidjs/solid#903

[Refs are not reactive](https://discord.com/channels/722131463138705510/722349143024205824/940407159999569970), so ref call does not need to be wrapped in the separate effect

The change I'm less confident with is that the ref now is called even if its value did not change - the ref call is moved above the `if (value === prev)` check, in both web (for spread only) and universal (always?) renderers. But this is necessary to have ref() called again after cleanup in the functional spread ( in `<span {...p()}>` in the last test )


ryansolid:
I see I will play with it. Generally looks good.

PR #124, opened 2022-04-13
feat: dynamic universal mode with dom optimizations and custom renderers

nksaraf:
This adds a new compiler mode, with `generate` option as `'dynamic'`. It allows us to create separate compilation outputs within the same file for different JSX elements. 

It is intended to be a more powerful version of the current `'universal'` mode and specifically benefits renderers that want to coexist with super efficient `dom` compiler. This allows those to not have to write their own renderers for `dom` and `svg` and opt in to the default compiler.

We can configure it as follows:
```typescript
import { HTMLElements, SVGElements } from "./elements";
let solidConfig = {
  moduleName: "solid-js/web",
  // @ts-ignore
  generate: "dynamic",
  renderers: [
    {
      name: "dom",
      moduleName: "solid-js/web",
      elements: [...HTMLElements, ...SVGElements],
    },
    {
      name: "universal",
      moduleName: "solid-three",
      elements: [],
    },
  ],
};

```
For any HTML and SVG elements, it will use `solid-js/web`'s imports and the `dom` compiler. For any remaining ones, it will use the "universal" compiler with import from "solid-three". 

We can include more that 2 renderers as well but you will have ensure that there is no confusion between which renderer would render an element. (We could get smarter about this in the future, where based on parent, we could disambiguate).


ryansolid:
Hell yes.. Thank you. This will probably need a minor bump to make versions clear. But this will be part of Solid 1.4 as well. Awesome work.

nksaraf:
Yeah there is a lot of stuff in here, let me know if you want me to walk you through the changes

PR #129, opened 2022-04-24
test: Add test for classList

PeterlitsZo:
In my opinion, adding test to make sure props `classList`, `class` and
`className` work well toghter is important.

WIP.

### DIFF
--- a/packages/dom-expressions/test/dom/classList.spec.jsx
+++ b/packages/dom-expressions/test/dom/classList.spec.jsx
@@ -64,4 +64,16 @@ describe("Test classList binding", () => {
       expect(div.className).toBe("px-4 py-2 text-sm");
     });
   });
+
+  test("With prop class and className", () => {
+    let div;
+    S.root(() => {
+      div = <div
+        className="px-1"
+        class="py-2"
+        classList={{ "text-sm": true, "danger": false }}
+      />;
+      expect(div.className).toBe("px-1 py-2 text-sm");
+    });
+  })
 });


PeterlitsZo:
Hello, I still have a question: Where did you define the `setProp`.

I know that if there is a component:

```
<div class="A" className="B" classList={{ "C": true }} />
```

It will turn into the code with `setProp`:

```
_$setProp(_el$, "class", "A");
_$setProp(_el$, "className", "B");
_$setProp(_el$, "classList", { C: true }); 
```

But I cannot find where is the function `setProp`. Could you please tell me?

ryansolid:
If I remember correctly `_$setProp` is only in the universal renderer and would be implemented by the creator of that renderer.

PeterlitsZo:
Oh I get it. So the `dom-expressions` will translate the source code into `_$setProp` and the render will define the function `setProp`.

So if I want to know how the library SolidJS deal with the props `class`, `className` and `classList`, I need to check the `setProp` in SolidJS.

PeterlitsZo:
I find the function `setProperty` in the folder `packages/solid/store/src/` of the project `soildjs/solid`. But the source code do not deal with `className` and `class`. `>_<`

PeterlitsZo:
:sob: sad.

ryansolid:
No. Dom Expressions has multiple outputs. The universal renderer has no idea of class because the target could be.. like a terminal or like WebGL. It's the only one that uses `_$setProp`. Don't worry about the universal output. 

The rest is handled by the DOM, SSR compilation and the client.js in dom-expressions.

PeterlitsZo:
Wow! Thanks! It is really cool... But where does the solid deal with the prop `class`, `className` and `classList`? I cannot find anything in the `dom-expressions`, neither the `solidjs`. :sob:

Any reply is helpful for me~

ryansolid:
This is the classList helper for DOM: https://dev.to/ryansolid/marko-for-sites-solid-for-apps-225e-temp-slug-7375806?preview=7dc4313047bb6f9bda82dafa64125293497de8a9bfe89715f96c805c6317e2cd25ddb3d9c6a4c2a6f0acf61133d1c6bcbb469089b6f1a1b0dab18ab4
The classList helper for SSR: https://dev.to/ryansolid/marko-for-sites-solid-for-apps-225e-temp-slug-7375806?preview=7dc4313047bb6f9bda82dafa64125293497de8a9bfe89715f96c805c6317e2cd25ddb3d9c6a4c2a6f0acf61133d1c6bcbb469089b6f1a1b0dab18ab4

Beyond that Babel JSX DOM Expressions does some special merging: https://github.com/ryansolid/dom-expressions/blob/main/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js#L222-L303 and https://github.com/ryansolid/dom-expressions/blob/main/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js#L222-L303. And some special logic in the server side spread: https://github.com/ryansolid/dom-expressions/blob/main/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js#L222-L303

I think that mostly covers it.

PeterlitsZo:
Thanks!

PR #135, opened 2022-06-09
fix *this* in events

mdynnl:
- fixes a infinite recursion bug from [this commit](https://github.com/ryansolid/dom-expressions/commit/cc3437dd25a4bf9fb94381d2dacde0469a673a2e?diff=split#diff-da2a5911aa8defd1b1e3649a3b811b7e67a2a7cea695d42b24a5e4ffd2b68ebdR90)
- implements `this` value in event handlers as per  https://html.spec.whatwg.org/multipage/webappapis.html#the-event-handler-processing-algorithm


ryansolid:
Thank you.

PR #136, opened 2022-06-14
Fix: Node process does not exit until timeout

fgnass:
Calling `renderToStringAsync` prevents Node.js from exiting until the timeout has elapsed. This PR fixes the issue by clearing the timeout once `asyncWrap` is resolved.

### DIFF
--- a/packages/dom-expressions/src/server.js
+++ b/packages/dom-expressions/src/server.js
@@ -45,9 +45,10 @@ export function renderToStringAsync(code, options = {}) {
       );
     }
   });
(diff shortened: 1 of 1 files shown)


edemaine:
FYI, an alternative is to [unref](https://nodejs.org/api/timers.html#timeoutunref) the timeout. But this PR's approach may be preferable because it doesn't rely on an internal NodeJS interface (e.g. for Deno).

ryansolid:
Thanks

PR #138, opened 2022-06-18
Warn on unnecessary rerenders

milomg:


ryansolid:
Hmm... trying to figure out if this is an annoying message or not. Haven't vetted enough projects to know if this is common place. What drew your attention to this?

modderme123:
I was using a resource to store and update an object, but that was triggering an update on all fields on an object (and I would imagine there might be a similar issue with signals in beginner's code). I realize that this could be annoying if intentional though

modderme123:
I was also wondering what this looks like on common projects so I tried it on the playground and there were no warnings, but that isn't a perfect test

ryansolid:
This is interesting. I think I might slip something like this in to try it. (Although looking at the PR there is an equality check that shortcuts this ever getting hit right above). But we could always lift it out. Anyway. This isn't getting merged so might as well keep things clean. Thanks.

PR #141, opened 2022-07-03
Treat `attribute={true}` like `attribute=""`

edemaine:
Fixes https://github.com/solidjs/solid/issues/1101 by making `<div foo={true}>` (and `<Div foo={true}>` behave like `<div foo>` and `<div foo="">` i.e. HTML Boolean attributes.

It would be good to add tests for this... but I could use help as I haven't touched the tests before.


ryansolid:
Thanks for the PR. Not going this direction though from the conclusion of the discussion. The decision was to keep the current behavior and update SSR to reflect it.

PR #144, opened 2022-07-05
WIP fix: untrack importNode calls for custom elements 

trusktr:
This avoids accidental infinite loops caused by constructors creating
and using reactive variables. This issue did not previously happen when
all templates used `cloneNode`. `v0.33.7` introduced `importNode` for
custom elements, and some programs started having infinite loops.


ryansolid:
I added this a while back and forgot to close.

PR #146, opened 2022-07-07
Handle special case attributes

### LINKED ISSUE #143: multiline attributes work in JSX templates, but not in `html` templates
JSX working: https://playground.solidjs.com/?hash=-570526116&version=1.4.1

`html` not working: https://playground.solidjs.com/?hash=382249257&version=1.4.1

It treats each line inside the multiline attribute as an attribute:

```html
<textarea one two three ...></textarea>
```

trusktr:
fixes #143

The PR has whitespace changes from prettier in the test file. Turn on "Hide whitespace" setting when viewing the diff.

- [x] switch to version number for `html-parse-string` package [once that is released](https://github.com/ryansolid/html-parse-string/pull/4), before merging this PR.

- [ ] update html-parse-string version one https://github.com/ryansolid/html-parse-string/pull/5 is merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ types/
 test/runtime.js
 packages/babel-plugin-jsx-dom-expressions/index.js
 !packages/babel-plugin-jsx-dom-expressions/types/
+package-lock.json
--- a/packages/lit-dom-expressions/package.json
+++ b/packages/lit-dom-expressions/package.json
@@ -15,7 +15,8 @@
   "scripts": {
     "build": "rollup -c && tsc",
     "test": "npm run build && jest",
-    "test:coverage": "npm run build && jest --coverage"
+    "test:coverage": "npm run build && jest --coverage",
+    "prepare": "npm run build"
   },
   "devDependencies": {
     "dom-expressions": "^0.33.13",
--- a/packages/lit-dom-expressions/package.json
+++ b/packages/lit-dom-expressions/package.json
@@ -19,6 +19,6 @@
   },
   "devDependencies": {
     "dom-expressions": "^0.33.13",
-    "html-parse-string": "0.0.6"
+    "html-parse-string": "trusktr/improve-attribute-parsing#d65d6a8aaa86321cc4d08c94d2fa055bc8792cfd"
   }
 }
--- a/packages/lit-dom-expressions/package.json
+++ b/packages/lit-dom-expressions/package.json
@@ -19,6 +19,6 @@
   },
   "devDependencies": {
     "dom-expressions": "^0.33.13",
-    "html-parse-string": "trusktr/improve-attribute-parsing#d65d6a8aaa86321cc4d08c94d2fa055bc8792cfd"
+    "html-parse-string": "^0.0.7"
   }
 }
--- a/packages/babel-plugin-jsx-dom-expressions/package.json
+++ b/packages/babel-plugin-jsx-dom-expressions/package.json
@@ -15,7 +15,8 @@
     "build": "rollup -c",
     "test": "npm run build && jest --no-cache",
     "test:coverage": "npm run build && jest --coverage --no-cache",
-    "prepublishOnly": "npm run build"
+    "prepublishOnly": "npm run build",
+    "prepare": "npm run build"
   },
   "dependencies": {
     "@babel/helper-module-imports": "7.16.0",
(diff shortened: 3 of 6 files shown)


trusktr:
The CI isn't able to install from git correctly. This can be tested locally by `npm link`ing `html-parse-string` into `lit-dom-expressions` before running `npm test`.

ryansolid:
Any reason to bring regexr in? I know we are talking bytes here but if we didn't need to mess with the regexp much probably isn't worth bringing.

trusktr:
I can remove it

The only reason was regexes being more readable/easy to compose. Initially I thought the problem was there, so it would've been easier to work with, but it ended up being in html-parse-string.

Been imagining a Babel plugin to compile ``` r`` ``` expressions to strings/regexps would be interesting...

trusktr:
PR updated

.gitignore
```diff
@@ -6,4 +6,5 @@ coverage/
 types/
 test/runtime.js
 packages/babel-plugin-jsx-dom-expressions/index.js
-!packages/babel-plugin-jsx-dom-expressions/types/
\ No newline at end of file
+!packages/babel-plugin-jsx-dom-expressions/types/
+package-lock.json
```
trusktr, line 10: This kept popping up, figured I'd ignore them since they aren't checked in (as it should be with libs)

packages/lit-dom-expressions/package.json
```diff
@@ -19,6 +19,6 @@
   },
   "devDependencies": {
     "dom-expressions": "^0.33.13",
-    "html-parse-string": "0.0.6"
+    "html-parse-string": "trusktr/improve-attribute-parsing#d65d6a8aaa86321cc4d08c94d2fa055bc8792cfd"
```
trusktr, line 22: - [ ] TODO, update this line to the new release

trusktr:
update html-parse-string, new tests now passing

trusktr:
I'm taking this for a spin in LUME right now

trusktr:
It seems to be that some of the cases that successfully pass through `html-parse-string` still fail after they pass through `lit-dom-expressions`.

For example, the following test,

```html
    /** @type {HTMLElement} */
    const el = html`
      <lume-box
        has="shader-material"
        receive-shadow="false"
        size="400 400 400"
        sidedness="double"
        uniforms='{
					"iTime": { "value": 0 },
					"iResolution": { "value": {"x": 0, "y": 0, "z": 0} }
				}'
        vertex-shader="
					varying vec2 vUv;

					void main() {
						vUv = uv;
						gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
					}
				"
        fragment-shader="
					#include <common>

					uniform vec3 iResolution;
					uniform float iTime;

					// The following is the default shader when you start a new shadertoy example.
					// By iq: https://www.shadertoy.com/user/iq
					// license: Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported License.

					// BEGIN SHADERTOY CODE {

					void mainImage( out vec4 fragColor, in vec2 fragCoord ) {
						// Normalized pixel coordinates (from 0 to 1)
						vec2 uv = fragCoord/iResolution.xy;

						// Time varying pixel color
						vec3 col = 0.5 + 0.5*cos(iTime+uv.xyx+vec3(0,2,4));

						// Output to screen
						fragColor = vec4(col,1.0);
						//fragColor = vec4(1.0, 0.3, 0.1, 1.0);
					}

					// END SHADERTOY CODE }

					varying vec2 vUv;

					void main() {
						mainImage(gl_FragColor, vUv / 2.0 * gl_FragCoord.xy);
					}
				"
      ></lume-box>
    `;

    expect(el.attributes.length).toBe(7);
```

fails with `20` instead of `7`, but as you can see in the following regexr the attributes are detected fine by the new html-parse-string regex:

https://regexr.com/6pc1b

So, now I need to figure out why `lit-dom-expressions` is further splitting the attributes.

trusktr:
Found it, and added it to the same PR I have open. Quote characters needed to be escaped into HTML entities: https://github.com/ryansolid/html-parse-string/pull/5/files#diff-d937d19143c8eb7c45d90735ae2a258a0a681496be1259451254beeb64a01fbfR7

ryansolid:
Awesome. Thank you.

trusktr:
@ryansolid the version of `html-parse-string` needs to be updated in this `package.json` still.

trusktr:
In my other comment I was wondering if you want to just bump `html-parse-string` to `0.1.0`, otherwise we have to also bump it here on every non-breaking change (semver treats all `0.0.x` bumps as breaking, but `0.1.x` bumps are treated as non-breaking patches).

PR #147, opened 2022-07-10
bump html-parse-string version

trusktr:
Note, we wouldn't need to bump the version for every non-breaking html-parse-string release if we bump it to `0.1.x` or higher.


trusktr:
Nvm, I see you did that already.

PR #150, opened 2022-07-11
Add `csstype`, fix CSSProperties

### LINKED ISSUE #142: Fix types in CSSProperties
[`interface CSSProperties`](https://github.com/ryansolid/dom-expressions/blob/main/packages/dom-expressions/src/jsx.d.ts#L284) is full of types like `CSSWideKeyword | any` which give some simple autocompletion but don't actually type check. The result is that `style` attributes are mostly unchecked. For example, it may appear that they accept functions wrapping reactive values, but they don't.  There's also the line `[propertyName: string]: any;` which fixes CSS variables and browser-specific overrides but also enables typos.

@LXSMNSYC suggested on Discord that [csstype](https://github.com/frenic/csstype) is a popular fix here, and we may be able to just drop it in as a replacement. But it might be a little strict; for example, I don't think CSS variables "just work" (see [overrides](https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors)) and [values don't yet support `!important`](https://github.com/frenic/csstype/issues/160).

Alternatively, a big improvement would be replacing `any` with `string`, `string | number`, etc. as appropriate. And the CSS variable/browser override rule could be ``[propertyName: `-${string}`]: any`` so it only applies to properties starting with `-`.

LXSMNSYC:
Fixes #142 

Additional:
Adds support for `style:*` and `class:*`, which allows you to do this
image

**NOTE** `csstype` is marked as a peer dependency. Solid must install this as a dependency (not dev dependency).


ryansolid:
Can we not do the additional for now? Those fields need to still be reviewed from a runtime perspective.

LXSMNSYC:
> Can we not do the additional for now? Those fields need to still be reviewed from a runtime perspective.

Sure. I've commented it out for now in case we want to include these set of features in the future. We can remove it anytime once we've decided.

PR #151, opened 2022-07-27
Add types for onAuxClick

edemaine:
As reported by @fabiospampinato, the types for `onAuxClick` were missing, despite the existence of [auxclick event](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event). He wrote in #typescript:

> I don't know why, but the middle click seems to fire an "auxclick" for me, not a "click", so this feels pretty much essential. A convenience onMiddleClick might be worth exploring too.


ryansolid:
Thanks

PR #153, opened 2022-07-30
Fix reconcileArrays assuming non-empty input array

EricNRodriguez:


ryansolid:
Is there a place where you hit this as a bug?

EricNRodriguez:
I was using it directly in a side project and came across it, thought I would correct it whilst I was here. I didn't find any cases where it caused a crash elsewhere, but I also didn't look extensively

ryansolid:
Ok the reason I asked is mostly it is a hot path and if it isn't neccessary Id skip. Plus we don't transpile and there are envs that don't support optional chaining so we'd need to make a slightly larger fix. Unless there is a bug/code path where this causes an issue Id be tempted to leave as is

EricNRodriguez:
Makes sense - I have no strong opinions so I'll close this off given the above info. Thanks Ryan

PR #157, opened 2022-08-06
feat: ungroup dynamics for universal renderer

vjoao:
### Summary

This PR removes the effect grouping for dynamics in the 'universal' generated code

### Description

A potential perf issue when complex objects that have reactive props inside them are passed as props in the template.

```ts
<element prop1={[ {x: 30}, { y: count() }]} prop2={{ key: anotherSignal() }} />
```

The transformer will group those 2 props into a single effect, which is neat, but because the values are complex, they will always fail an equality check:

```ts
_$effect(_p$ => {
  const _v$ = [{
    x: 30
  }, {
    y: count()
  }],
  _v$2 = {
    key: anotherSignal()
  };
  _v$ !== _p$._v$ && (_p$._v$ = _$setProp(_el$, "prop1", _v$, _p$._v$));
  _v$2 !== _p$._v$2 && (_p$._v$2 = _$setProp(_el$, "prop2", _v$2, _p$._v$2));
  return _p$;
}, {
  _v$: undefined,
  _v$2: undefined
});
```

Note that _v$ will be a new array on every effect run. _v$2 will be a new object on every effect run. Both will fail the equality check and reassign on any signal change of that effect, be it count() or another().

Removing grouping means you trade memory for execution speed because now dynamics only execute for the props they depend on. If you want this to be a configuration option let me know.

### Testing

Unit tests green

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/universal/template.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/universal/template.js
@@ -39,59 +39,19 @@ function wrapDynamics(path, dynamics) {
   const config = getConfig(path);
 
   const effectWrapperId = registerImportMethod(path, config.effectWrapper);
+  const prevValue = t.identifier("_$p");
 
-  if (dynamics.length === 1) {
-    const prevValue = t.identifier("_$p");
-
+  return dynamics.map(({ elem, key, value }) => {
     return t.expressionStatement(
       t.callExpression(effectWrapperId, [
         t.arrowFunctionExpression(
           [prevValue],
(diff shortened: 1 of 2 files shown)


ryansolid:
Sorry for the delay on this. I'm weighing doing it this way versus bigger changes we could make and I'd rather not change and change again. Mostly looking if we can change some more around props.

victor-wm:
No worries @ryansolid - take your time. Also, feel free to use me as a test subject for anything you need around this.

ryansolid:
Stale (2022). Not taking a universal effect-ungroup onto the Solid 1.x maintenance line. The 2.x compilers live in the Solid repo now.

PR #160, opened 2022-08-18
Fix undefined variable in SSR spread

jon-hall:
Had this show up in a build on my own project the other day (`ReferenceError: results is not defined`) and it looked easy enough to fix, I also added a simple test case to recreate the error/verify the fix.


ryansolid:
Thanks. No clue how I missed that but thank you.

PR #161, opened 2022-08-27
fix(type): style use variable type error

zoy-l:
- fix: https://github.com/solidjs/solid/issues/1184 

can also use more strict 
```{ [key: `—${string}`]: string | number | undefined }```


ryansolid:
Thanks

zoy-l:
sorry, I don't know when { [key: `—${string}`]: string | number | undefined } is escaped, use `--`

PR #176, opened 2022-11-04
fix spread

mdynnl:
might fix https://github.com/solidjs/solid/issues/1330


mdynnl:
Maybe there's more to handle (hydratable tests are failing too) but that's beyond my limit.

PR #178, opened 2022-11-15
add CSS Typed OM types

MrFoxPro:
https://github.com/frenic/csstype/issues/170

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dom-expressions-build",
-  "version": "0.31.0",
+  "version": "0.35.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


MrFoxPro:
I'm not sure if `import '@types/w3c-css-typed-object-model-level-1';` is required. 
Seems like it works without import.

lxsmnsyc:
Copying my message from Discord

I cannot approve it for the following reasons:
- The specification is still yet to be standardized (it's still in level 1, and is marked experimental)
- The values themselves are browser-only and is highly incompatible with SSR
- This requires changing the runtime for serialization and application purposes. CSS Typed OM is applied differently from a normal string property.

ryansolid:
I'm going to go with @lxsmnsyc judgment for now. But willing to re-examine in the future.

PR #186, opened 2022-12-19
fix-dom-function-children

orenelbaum:
Fixes a problem where passing functions (either arrow functions or function expressions) with a block body would break compilation.
Fixes https://github.com/solidjs/solid/issues/1423


ryansolid:
Thank you.

PR #187, opened 2022-12-24
fix-escaping-booleans

orenelbaum:
Fixes https://github.com/solidjs/solid/issues/1443
Please review this PR since I'm not 100% confident about it.

### DIFF
--- a/packages/dom-expressions/src/server.js
+++ b/packages/dom-expressions/src/server.js
@@ -319,7 +319,16 @@ export function ssrElement(tag, props, children, needsId) {
 }
 
 export function ssrAttribute(key, value, isBoolean) {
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah we don't want to change the behavior of escaping. The problem is in the JSX transform I think. We should include the boolean logic in the escaping not outside. 

Right now we get like:
```js
_$ssr(_tmpl$, _$ssrHydrationKey(), !!_$escape(count()))
```

Instead of:
```js
_$ssr(_tmpl$, _$ssrHydrationKey(), _$escape(!!count()))
```

ryansolid:
I've submitted the fix to compilation. So this is not needed. Thanks though.

PR #191, opened 2022-12-31
Reorganize jsx-runtime html/svg type definitions 

obask:
The idea behind was PR was originally in figuring out where `jsx-runtime` definitions are coming from and apply minimal changes without breaking someones IDE or existing projects. And also adding jsdoc like references to where each element comes from.
Added support template html field. Added `verify_types.ts` file to run extra static checks in an IDE.

This files will override definitions in the solid repository.
Tested with a solid/web tsx files, but I had to temporally change tsconfig.path to make it see this files.

### DIFF
--- a/packages/dom-expressions/src/jsx-h.d.ts
+++ b/packages/dom-expressions/src/jsx-h.d.ts
@@ -99,100 +99,24 @@ export namespace JSX {
   type OnCaptureAttributes<T> = {
     [Key in keyof CustomCaptureEvents as `oncapture:${Key}`]?: EventHandler<T, CustomCaptureEvents[Key]>;
   }
-  interface DOMAttributes<T> extends CustomAttributes<T>, DirectiveAttributes, DirectiveFunctionAttributes<T>, PropAttributes, AttrAttributes, OnAttributes<T>, OnCaptureAttributes<T> {
+  interface DOMAttributes<T> extends CustomAttributes<T>, DirectiveAttributes, DirectiveFunctionAttributes<T>,
+                                     PropAttributes, AttrAttributes, OnAttributes<T>, OnCaptureAttributes<T>,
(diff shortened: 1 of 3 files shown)


ryansolid:
Is there a reason for how events are divided? I see `focusout` and `dragexit` not included in the new type but `focus` and `dragenter` is?

obask:
I concluded that these HTML elements specs are taken from  https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L5814 file. If you `go to a definition` of GlobalEventHandlers I've added in a jdoc comment, you'll see that `onfocusout` and `ondragexit` aren't derived from this class. So they are added via some other class in Solid's version of `jsx-runtime`. I noticed that the most libraries customize own implementation for the fields they want users to use, but I couldn't find much documentation about it.

React version of `ondragexit`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1496
Preact is mostly compatible, but it has it's own flavor: https://github.com/preactjs/preact/blob/master/src/jsx.d.ts#L589

So theoretically we can replace all of that with this snippet:
```
    type HTMLAttributes<RefType extends HTMLElement> = Partial<RefType> | {
        class?: string
    }
    type IntrinsicElementsType =  {
        [K in keyof HTMLElementTagNameMap]: HTMLAttributes<HTMLElementTagNameMap[K]>
    }
    export interface IntrinsicElements extends IntrinsicElementsType {}
```
But it don't give much flexibility and ability to add camel case fields(it also breaks Webstorm code navigation and makes VSCode navigation less intuitive).

PS let me know if you think this are reasonable assumptions or it's better to abandon this change in favor(or in addition) to figuring out issues related to `tsconfig.json`.

ryansolid:
Ok makes sense. Thanks.

PR #194, opened 2023-01-08
Fix test failures when using pnpm and run npm-upgrade on all dependencies

obask:
#### Generating `pnpm-lock.yaml` and fixing issues with whitespaces
The initial intent was to just fix `pnpm i && pnpm build && pnpm test` command and reduce confusion when tests suddenly test fail with mismatched white spaces when someone is trying to run them with `pnpm`(since there was no `pnpm-lock.yaml` which results in some new pretty-formatter version). 
I haven't figured out how to fix this without updating test fixtures, so the first commit contains both. I think the babel plugin-tester is suppose to apply the same formatting to transform results and `output.js`, but it wasn't doing this for some reason).

#### [optional] `npx npm-upgrade`
The second change is a nice-to-have feature. It involves updating package versions to their latest using `npm-upgrade`. However this resulted in some additional tweaking such as adding a `rollup --bundleConfigAsCjs ` flag and setting `jsdom` test environment.

#### Testing
Tested with the `solid` project (and solid todo app template) using `pnpm i && pnpm link ../dom-expressions/packages/*` (somehow it drops `babel-plugin-transform-rename-import` dependency if I run `pnpm link` first; I don't think it's related, but it would be nice to add it explicitly).

#### dom-expressions tests output:
Screenshot 2023-01-07 at 6 52 41 PM

#### solid  tests output:
Screenshot 2023-01-07 at 6 44 57 PM


obask:
It looks like `lerna` assumes that if you have `pnpm-lock.yaml` you should be using pnpm... so it looks like it is a "one way door". The only bright side is that the projects build scripts look similar to `solidjs/solid` now.

ryansolid:
Yeah I hit this when I tried to update a few weeks back and just pushed it off at the time. Thank you for taking the time to address this.

PR #200, opened 2023-02-01
Add void elements check for `ssrElement`

lxsmnsyc:


packages/dom-expressions/src/server.js
```diff
@@ -4,6 +4,8 @@ import stringify from "./serializer";
 export { stringify };
 export { createComponent } from "rxcore";
 
+// Based on https://github.com/WebReflection/domtagger/blob/master/esm/sanitizer.js
+const VOID_ELEMENTS = /^(?:area|base|br|col|embed|hr|img|input|keygen|link|menuitem|meta|param|source|track|wbr)$/i;
```
lxsmnsyc, line 8: This was carried over from [`lit-dom-expressions`](https://github.com/ryansolid/dom-expressions/blob/main/packages/lit-dom-expressions/src/index.ts#L35)

packages/dom-expressions/src/server.js
```diff
@@ -285,14 +287,15 @@ export function ssrStyle(value) {
 
 export function ssrElement(tag, props, children, needsId) {
   let result = `<${tag}${needsId ? ssrHydrationKey() : ""} `;
+  const skipChildren = VOID_ELEMENTS.test(tag);
```
lxsmnsyc, line 290: Early test so we can re-use for two cases:
- Checking for child-related properties
- Markup result (for self-closing)

PR #205, opened 2023-02-16
refactor: keep persistent type

ulivz:
# Motivation

We should not modify the type of a variable at runtime, this can be error prone and misleading:

```ts
results.decl.unshift(decl);
results.decl = t.variableDeclaration("const", results.decl); // decl's type changed from `array` to `t.VariableDeclaration`
```

To fix this, we can create another property to save the original declarations.


ryansolid:
Thank you

PR #208, opened 2023-02-25
fix (docs): updated "compiles to" code example for babel-plugin-jsx-dom-expressions

i-Welch:
**Reasoning**
- based on this example I was pretty confused when playing around with certain playground implementations as they generally had different output then I'd expect based on the example here. After doing a minimal repo to try and see why the code examples differ I found that the docs didn't seems correct anymore (either due to implementation drift or due to the default configurations not being proper to get the documented output) If the case is the former then I think we should update it. If the case is the latter we should add the configs to the README.md example (and I'd be happy to do that).

**To Test**
- I'd encourage anyone to test a blank project against the example and make sure they get the same output as I did.

**Place for feedback**
-
- Not sure if the `_$delegateEvents(["click"])` piece should be included in the output. Open to input there.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/README.md
+++ b/packages/babel-plugin-jsx-dom-expressions/README.md
@@ -38,29 +38,33 @@ const view = ({ item }) => {
 Compiles to:
 
 ```jsx
-import { insert as _$insert } from "dom";
+import { template as _$template } from "dom";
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks. I honestly probably haven't touched that part of the Readme in years.

PR #209, opened 2023-03-01
fix: add href attribute to GradientElementSVGAttributes interface

emdede:
There might be other types missing, I just stumbled on this in my project. See [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href)

I am uncertain about the implications of `FunctionMaybe` type but I followed example of similar attribute typing.


ryansolid:
It's because of how we do HyperScript. It's all good.

PR #217, opened 2023-03-11
Fix server.js error missing `r`

lxsmnsyc:
Recent error caused all `renderToString` calls to be broken (this heavily affected Astro users)


ryansolid:
Thanks. I changed this from my phone before seeing the PR. I will fix this up when I get home.

PR #220, opened 2023-03-14
fix: pathLength added to ShapeElementSVGAttributes

Ladvace:
**Summary:**

I've added the missing `pathLength` to the `ShapeElementSVGAttributes` interface

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pathLength


ryansolid:
Thanks

PR #225, opened 2023-03-24
Supported non-standard input attributes

elite174:
Referring to this issue https://github.com/solidjs/solid/issues/1642 I added non-standard attributes (see this https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#non-standard_attributes).

### DIFF
--- a/packages/dom-expressions/src/jsx.d.ts
+++ b/packages/dom-expressions/src/jsx.d.ts
@@ -798,6 +798,7 @@ export namespace JSX {
     accept?: string;
     alt?: string;
     autocomplete?: string;
+    autocorrect?: "on" | "off";
     autofocus?: boolean;
     capture?: boolean | string;
     checked?: boolean;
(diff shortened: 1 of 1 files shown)


ryansolid:
Thanks

jesseb34r:
I'm definitely happy to have the non-standard search attributes added. If we are going to add these, should we add all of the non-standard input attributes? [mdn reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#non-standard_attributes:~:text=image%20input%20type.-,Non%2Dstandard%20attributes,-The%20following%20non)

You added `autocorrect`, `incremental`, and `results`. There is also `autocapitalize`, `orient`, and `webkitdirectory`. I don't know how useful they are to have on the type but they are valid attributes.

elite174:
@jesseb34r thank you! I guess I missed them. Will make a PR today!

PR #227, opened 2023-03-27
Add `loading` JSX attribute type to `<iframe>`

kidonng:
Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#loading

From https://github.com/solidjs/solid/pull/1652


ryansolid:
Thank you

PR #231, opened 2023-04-03
add conditions to the style

sirenkovladd:
if the variable is null or undefined then remove this style

https://github.com/solidjs/solid/issues/1299


ryansolid:
Thank you

PR #233, opened 2023-04-04
add support syntax use:{function}={expression}

sirenkovladd:
Fixed
https://github.com/solidjs/solid/issues/1666

Add testing of dynamic functions in tests


ryansolid:
Thank you. This will get merged now. I will consider it experimental until 1.8. But we can start testing it.

PR #234, opened 2023-04-06
Fix class:foo={undefined}/classList={{foo: undefined}} bugs

### LINKED ISSUE #133: class:foo={undefined} doesn't cast to boolean
I was reviewing https://github.com/solidjs/solid-docs/commit/4850acf52752db5ebe28e05f593567339d0b0c83 but don't understand the code well enough to read it, so I just tested it.

Bug: `<div class:foo={undefined}/>` seems to compile to `_el$2.classList.toggle("foo", undefined);`, which is the opposite of what we want (it turns the class on).  Similarly `<div class:foo={foo}/>` where `foo` is a local variable. So I guess the literals not matching the `if` need to be `!!`d at compile time.

By contrast, `<div class:foo={foo()}/>` compiles correctly to
```js
_v$3 = !!foo();
...
_v$3 !== _p$._v$3 && _el$2.classList.toggle("foo", _p$._v$3 = _v$3);
```
Also `classList` works great.

nicketson:
This should fix #133.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -177,7 +177,10 @@ export function setAttr(path, elem, name, value, { isSVG, dynamic, prevId, isCE,
         t.memberExpression(elem, t.identifier("classList")),
         t.identifier("toggle")
       ),
-      [t.stringLiteral(name), value]
+      [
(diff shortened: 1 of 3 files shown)


ryansolid:
Thank you

nicketson:
> Thank you

No problem! BTW, the `class:foo={blah}` syntax is very nice.

PR #238, opened 2023-04-10
Bump seroval to 0.7.0

lxsmnsyc:
This is just a patch.

seroval 0.5.0 had some broken serialization-related stuff but mainly on Typed arrays. seroval has also added support for 8 new types. There are no breaking changes though.


ryansolid:
Thanks

PR #239, opened 2023-04-16
Add some JSX attribute types

kidonng:
- [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) global attribute
- [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) global attribute: `until-found` (new) and `hidden` (missing) value


ryansolid:
Thanks

PR #240, opened 2023-04-17
Shave bytes by removing quotes and spacing from attributes

intrnl:
This PR attempts to shave more bytes off the template code by removing or replacing the quotes, along with removing the spacing between them when it's next to a quoted attribute value.

The code would eagerly use single quotes when it can't remove the quotes entirely, this is mostly because the template literals would usually be converted into regular string by the transformer/minifier, where they would normally be double quotes. I think by doing this, gzip compression would atleast see that there are more repetitions with `("`, but that's something that has yet to be tested.

The one thing I haven't done here is to adjust `escapeHTML` code, as it still assumes that it's being wrapped in a double quote. I wasn't sure how to deal with it, so I just kept it as is but it'd definitely have to be changed I think.

Alternatively we could just drop the code relating to switching between double and single quotes for now.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/attributeExpressions/output.js
+++ b/packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/attributeExpressions/output.js
@@ -11,27 +11,25 @@ import { classList as _$classList } from "r-dom";
 import { use as _$use } from "r-dom";
 import { spread as _$spread } from "r-dom";
 import { mergeProps as _$mergeProps } from "r-dom";
-const _tmpl$ = /*#__PURE__*/ _$template(
-    `<div id="main"><h1 class="base" id="my-h1"><a href="/">Welcome`
-  ),
+const _tmpl$ = /*#__PURE__*/ _$template(`<div id=main><h1 class=base id=my-h1><a href=/>Welcome`),
   _tmpl$2 = /*#__PURE__*/ _$template(`<div><div></div><div> </div><div>`),
   _tmpl$3 = /*#__PURE__*/ _$template(`<div foo>`),
   _tmpl$4 = /*#__PURE__*/ _$template(`<div>`),
-  _tmpl$5 = /*#__PURE__*/ _$template(`<div class="a b">`),
-  _tmpl$6 = /*#__PURE__*/ _$template(`<input type="checkbox">`),
-  _tmpl$7 = /*#__PURE__*/ _$template(`<div class="\`a">\`$\``),
-  _tmpl$8 = /*#__PURE__*/ _$template(`<button class="static hi" type="button">Write`),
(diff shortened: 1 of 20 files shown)


ryansolid:
Nice. Yeah we did this with Marko always intended to look into it.

packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/attributeExpressions/output.js
```diff
@@ -11,27 +11,25 @@ import { classList as _$classList } from "r-dom";
...
-  _tmpl$10 = /*#__PURE__*/ _$template(`<div class="bg-red-500 flex flex-col">`),
-  _tmpl$11 = /*#__PURE__*/ _$template(`<div><input readonly=""><input>`),
-  _tmpl$12 = /*#__PURE__*/ _$template(`<div data="&quot;hi&quot;" data2="&quot;">`),
+  _tmpl$5 = /*#__PURE__*/ _$template(`<div class='a b'>`),
+  _tmpl$6 = /*#__PURE__*/ _$template(`<input type=checkbox>`),
+  _tmpl$7 = /*#__PURE__*/ _$template(`<div class='\`a'>\`$\``),
+  _tmpl$8 = /*#__PURE__*/ _$template(`<button class='static hi'type=button>Write`),
+  _tmpl$9 = /*#__PURE__*/ _$template(`<button class='a b c'>Hi`),
+  _tmpl$10 = /*#__PURE__*/ _$template(`<div class='bg-red-500 flex flex-col'>`),
+  _tmpl$11 = /*#__PURE__*/ _$template(`<div><input readonly=><input>`),
```
lxsmnsyc, line 24: I believe this is a mistake. What happens if there's `<input type="" readonly="">`

intrnl: Can't easily check at the moment and I haven't exactly looked at this PR in quite a while but I believe that would work fine, if a non-quoted attribute meets a whitespace or a delimiter like > then it would just consider <empty string> to be its value

lxsmnsyc: I mean the trailing equals works, what I'm saying is we don't have this specific test so I don't know what happens to the example code I gave, because I'm thinking it may spit out `<input type= readonly=>` which is most likely invalid.

ryansolid: It does end up being invalid. It's fine. I realized we were escaping stuff a certain way anyway. I keep the spirit of this and just went with current double quote stuff without doing anything too smart with the trailing "" scenario. We can revisit it if we want to later.

PR #242, opened 2023-04-26
Fix JSX missing type in webstorm

NoDocCat:
The webstorm error: [JSX: Completion not triggered by JSX.IntrinsicElements declared by "extends"](https://youtrack.jetbrains.com/issue/WEB-56801/JSX-Completion-not-triggered-by-JSX.IntrinsicElements-declared-by-extends)

Reference: [@types/react](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8507d1ec567df3fbac1e896ec97a260cccc5910d/types/react/index.d.ts#L3168)

### DIFF
--- a/packages/dom-expressions/src/jsx.d.ts
+++ b/packages/dom-expressions/src/jsx.d.ts
@@ -1940,10 +1940,18 @@ export namespace JSX {
       ZoomAndPanSVGAttributes {
     viewTarget?: string;
   }
+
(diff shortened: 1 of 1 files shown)


ryansolid:
I see. that's unfortunate. Types were cleaner before, but I suppose Webstorm breaking on it is unacceptable.

NoDocCat:
Maybe publish a npm package to override the type declaration is a good choice.

PR #244, opened 2023-05-02
Make `elementtiming` and  `fetchpriority` optional properties

yhdgms1:
I see there was an [Issue](https://github.com/solidjs/solid/issues/1690) about missing properties, but I suppose they should be optional, because if I do 

```ts
import type { JSX } from 'solid-js';

type ImgAttrs = JSX.ImgHTMLAttributes<HTMLImageElement>;

const defaultProps = {
  src: 'http://'
} satisfies ImgAttrs;
```

I get a error that `elementtiming` and  `fetchpriority` is required


bniwredyc:
:shipit:

usmanyunusov:
:(

PR #246, opened 2023-05-02
add heading tags to const DOMElements

bigmistqke:
As mentioned [here](https://github.com/solidjs/solid/issues/1713) heading tags are currently not included in DOMElements. This pr adds h1-h6 to the constant.

### DIFF
--- a/packages/dom-expressions/src/constants.js
+++ b/packages/dom-expressions/src/constants.js
@@ -469,7 +469,13 @@ const DOMElements = /*#__PURE__*/ new Set([
   "video",
   "wbr",
   "xmp",
-  "input"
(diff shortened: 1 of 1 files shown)


nksaraf:
Looks good to me. My bad on missing those!

bigmistqke:
Lovely :-)

I probably should have checked if there were other valid html-elements missing in the list. It's a bit tricky since valid html-tags are a bit of a moving target, and maybe it shouldn't be the responsibility of `dom-expression` to keep track of the spec? There is a package called [html-tags](https://www.npmjs.com/package/html-tags] which we could depend on to do this work for us.

PR #249, opened 2023-05-12
updates `draggable` attribute to allow `"true" | "false"`

titoBouzout:
`draggable="false"` was giving me a typescript error on the JSX of the editor. 

I have a question, all the booleans used on attributes on this file can be updated to be `boolean | "true" | "false"` right? (besides any other allowed value). If so, I can take a look to update them all.


ryansolid:
The psuedo-booleans like this should have string "true"/"false" allowed. But not all booleans. Just the ones like draggable that take strings I believe. We should actually make sure "false" does what we thing it does in a context that can't be compiled optimized out. Typical boolean attributes that are always true if present shouldn't allow for this though.

PR #253, opened 2023-05-24
feat: allow use documents other than window.document during render

pedro00dk:
Please check: https://github.com/solidjs/solid/pull/1740
It has the description for this PR too.


ryansolid:
Not landing on the 1.x line. The gap is still real (scoped registries / shadow-root documents) but a `document` field on the owner was the wrong layer — see https://github.com/solidjs/solid/pull/1740.

Moved to Solid 2 as an enhancement: https://github.com/solidjs/solid/issues/3056

Issue #247, opened 2023-05-05
Can't use `this` on prop function


titoBouzout:
I was trying to use `this` on a function given to props. It gets changed to _self$. Possibly this has to do with component classes from mobx-jsx? 

```js
function Component() {
  return (
    <div
      prop:update={function () {
        console.log(this);
      }.bind({ one: true })}
    />
  );
}
```

compiles to:

```js
function Component() {
  const _self$ = this;
  return (() => {
    const _el$ = _tmpl$();
    _$effect(() => _el$.update = function () {
      console.log(_self$);
    }.bind({
      one: true
    }));
    return _el$;
  })();
}
```
Playground: https://playground.solidjs.com/anonymous/393c3426-bea9-4d4c-8af9-5cc28d451a47

titoBouzout:
fixed in https://github.com/ryansolid/dom-expressions/commit/5b7b53d2ecb1c4fea679fbae6d165451fb5c3393

PR #256, opened 2023-06-01
Comments out transformation of `this`

titoBouzout:
Related to https://github.com/ryansolid/dom-expressions/issues/247

It comments out the transformation of `this` to `$self` as its breaking function binding in solid. 
Presumably the transformation is for compatibility with `mobx-jsx` which I guess is barely used compared to solid.


ryansolid:
Not sure what to do with this one. It's a bit of tug of war. It originally came about because of people doing web components with Solid. I have basically no use for `this` at all but it was causing issues for people.

PR #257, opened 2023-06-04
fix: possible fix for hydration bug where it was killing text nodes o…

i-Welch:
**Work Done**
-
- There is a bug when doing hydration where it seems to kill text nodes on inital hydration if they are wrapped in a signal at some point.
- Recreation here https://playground.solidjs.com/anonymous/ecb652e0-66e6-4289-b1f0-381ed1067fec (this playground does a few weird things to simulate hydration but I have a test project I can also send which exhibits the same behavior)
- Note: this reverts a performance optimization from a while ago and may not really be addressing the root issue which is that the hydration step if manipulating the DOM at all (or at least its manipulating the dom in such a way that this bug has occurred as a side effect)

**To Test**
-
- Create a simple hydration app based on the playground above and test this version of dom-expressions vs main and note that the bug is remediated. If you need help testing or something feel free to reach out

**To Do**
-
- I'm going to be working on a different fix that just stop hydration from manipulating the dom (or at least keeps it from nuking text nodes) but that may be a bit more complicated.

### DIFF
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -398,7 +398,6 @@ function insertExpression(parent, value, current, marker, unwrapArray) {
   parent = (multi && current[0] && current[0].parentNode) || parent;
 
   if (t === "string" || t === "number") {
-    if (sharedConfig.context) return current;
     if (t === "number") value = value.toString();
     if (multi) {
       let node = current[0];
--- a/packages/dom-expressions/test/ssr/hydrate.spec.js
+++ b/packages/dom-expressions/test/ssr/hydrate.spec.js
@@ -13,6 +13,35 @@ describe("r.hydrate", () => {
     _tmpl$3 = r.template(`<div>Last</div>`);
   let result, rendered;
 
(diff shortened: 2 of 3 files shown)


i-Welch:
I've tried to see if it's possible to just not insert the generated dom into root element but that seems to just break reactivity generally since the hydration code still attaches all reactivity to the generated nodes and not the existing nodes in the DOM

i-Welch:
@ryansolid pinging this. This seems to be a pretty good fix at the cost of the performance improvement. If you don't think that's an acceptable outcome or I'm mistaken on what that code I removed actually does I'm open to spending more time here to try and find a different fix (which I think I have a pretty good idea what the fix will be but it's a bit more complicated) I'm good with that I'd just like a bit of movement in this.

ryansolid:
The examples here are creating new nodes rather than hydrating on the client (including in the test). Ie it isn't equivalent to what Solid's compiler generates. I don't see us merging this as the fix so I'm closing this. I need a better reproduction.

PR #259, opened 2023-06-11
feat: camelCase CSS properties in style objects

ginnwork:
This PR adds the ability to use camelCase CSS property names in style objects.

### Example

```jsx
// old
<div style={{ 'margin-top': '40px' }}>
```

```jsx
// new
<div style={{ marginTop: '40px' }}>
```

### Testing

Testing has been added to ensure the Babel transform converts camelCase CSS property names to kebab-case ones.

### Definitions

The JSX definitions file has also been updated and TypeScript/IntelliSense recognizes the camelCase CSS property names.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -56,6 +56,8 @@ const alwaysClose = [
   "fieldset"
 ];
 
+const upperCaseChars = /[A-Z]/g;
+
 export function transformElement(path, info) {
   let tagName = getTagName(path.node),
     config = getConfig(path),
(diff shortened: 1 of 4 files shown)


ryansolid:
The challenge with these sort of changes is they have to be applied both at compile and at runtime to handle things like variables being passed in or spreads. So changing in compiler isn't enough. It's part of the way I have done this. I think having a single way to set stuff is better generally as it doesn't get weird conflicts say in spreads. I guess you could always convert first, but things get more complex when you have a combination of spread and non-spread. There may be a time to address this in the future but spreads make for a lot more complexity here.

edemaine:
I was going to suggest warnings about camel case in dev mode, but it turns out this is already a [eslint-plugin-solid lint rule](https://github.com/solidjs-community/eslint-plugin-solid/blob/main/docs/style-prop.md). There's even a `--fix` option to fix mistaken usage of camel case (e.g. coming from React). I imagine it only covers easy cases (not dynamic objects) but hopefully it catches most misuse.

ryansolid:
Yeah I don't see us changing this. Directionally we've been looking how to make things work a single way instead of multiple because it makes composition a pain.

PR #261, opened 2023-06-12
fix names of `http-equiv` and `accept-charset` attributes

### LINKED ISSUE #260: Incorrect TypeScript definition for `http-equiv` and `accept-charset` attributes
The `http-equiv` attribute on `<meta>` element (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv) and `accept-charset` attribute on `<form>` element (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#accept-charset) are the only two exceptions in all HTML attributes that they use kebab case.

dom-expression (and Solid) doesn't have any special handing for them (as opposite to React who accepts camelCase but emits kebab case), but the TypeScript definition for these two attributes are in camelCase and lower case:

https://github.com/ryansolid/dom-expressions/blob/a85d688353fef165b0456d5e192f4359dc00d5cb/packages/dom-expressions/src/jsx.d.ts#L989-L995

https://github.com/ryansolid/dom-expressions/blob/a85d688353fef165b0456d5e192f4359dc00d5cb/packages/dom-expressions/src/jsx.d.ts#L841-L853

yume-chan:
Fixes #260

### DIFF
--- a/packages/dom-expressions/src/jsx-h.d.ts
+++ b/packages/dom-expressions/src/jsx-h.d.ts
@@ -741,7 +741,7 @@ export namespace JSX {
     name?: FunctionMaybe<string>;
   }
   interface FormHTMLAttributes<T> extends HTMLAttributes<T> {
-    acceptcharset?: FunctionMaybe<string>;
+    "accept-charset"?: FunctionMaybe<string>;
     action?: FunctionMaybe<string>;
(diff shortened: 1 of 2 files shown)


ryansolid:
Yeah not many native attributes have dashes (outside of SVG) but those do and should reflect that. Thank you.

PR #264, opened 2023-06-29
Add enterkeyhint to textarea type def

SamJBerry:
Already exists on` <input>` but not on `<textarea>`. 

https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-enterkeyhint-attribute


ryansolid:
Thanks

PR #272, opened 2023-08-21
Unwrap arrow function IIFEs in component props

intrnl:
As per pull request title, unwraps arrow function IIFEs if they are directly used within a component prop.

the following `<Show />` pattern can be quite common, and this saves up some of the bytes incurred by it
```jsx
<Show
  when={(() => {
    const val = value();
    if ("key" in val) {
      return val;
    }
  })()}
>
  {(val) => (
    <>
      ...
    </>
  )}
</Show>
```

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/component.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/component.js
@@ -141,22 +141,44 @@ export default function transformComponent(path) {
               checkTags: true
             })
           ) {
-            let expr =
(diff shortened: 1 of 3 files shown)


intrnl:
`this` usage should be a non-issue as arrow functions `this` has already been handled by the plugin, but we can make an additional test case for this

ryansolid:
I like this. It does add complication to the compiler but it is definitely the type of thing we can do. I'm trying to think of anywhere this falls down but I don't think so.

PR #275, opened 2023-09-18
feat: streaming serializer

lxsmnsyc:
This is a work-in-progress. This PR integrates seroval's new streaming serializer to be utilized by SolidJS as a whole. With the new serializer, SolidJS will be able to:
- Dedupe references in both nested objects and across resources
- Allows SolidJS to serialize `Promise` and `ReadableStream` instances 
- Allow nested `Promise` instances to be naturally deferred (only top-level `Promise` instances can be blocked). These instances are then serialized and streamed later on.
- Part of the HydrationScript would be delegated to seroval as seroval has its own "hydration" script.

### DIFF
--- a/packages/dom-expressions/package.json
+++ b/packages/dom-expressions/package.json
@@ -20,11 +20,11 @@
   },
   "peerDependencies": {
     "csstype": "^3.0",
-    "seroval": "^0.7.0"
+    "seroval": "^0.10.1"
   },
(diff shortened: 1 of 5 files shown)


ryansolid:
Hey @lxsmnsyc lets just give you repo access and I'm going to merge this into `next` so we can build this out together.

PR #276, opened 2023-09-25
Reduce codegen size for dynamic expressions

intrnl:
This pull request attempts to reduce the size of the code emitted from merging the dynamic expressions together into one effect.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
@@ -129,65 +129,55 @@ function wrapDynamics(path, dynamics) {
       ])
     );
   }
-  const declarations = [],
-    statements = [],
-    identifiers = [],
-    prevId = t.identifier("_p$");
+
+  const statements = [];
+
   dynamics.forEach(({ elem, key, value, isSVG, isCE, tagName }) => {
     const identifier = path.scope.generateUidIdentifier("v$");
+    path.scope.push({ kind: "let", id: identifier });
+
+    if (key === "classList" || key === "style") {
+      const statement = t.expressionStatement(
+        t.assignmentExpression(
+          "=",
+          identifier,
+          setAttr(path, elem, key, value, {
+            isSVG,
+            isCE,
+            tagName,
+            dynamic: true,
+            prevId: identifier,
+          }),
+        ),
+      );
+
+      statements.push(statement);
+      return;
+    }
+
     if (key.startsWith("class:") && !t.isBooleanLiteral(value) && !t.isUnaryExpression(value)) {
       value = t.unaryExpression("!", t.unaryExpression("!", value));
     }
-    identifiers.push(identifier);
-    declarations.push(t.variableDeclarator(identifier, value));
-    if (key === "classList" || key === "style") {
-      const prev = t.memberExpression(prevId, identifier);
-      statements.push(
+
+    const statement = t.ifStatement(
+      t.binaryExpression("!==", identifier, t.assignmentExpression("=", identifier, value)),
+      t.blockStatement([
         t.expressionStatement(
-          t.assignmentExpression(
-            "=",
-            prev,
-            setAttr(path, elem, key, identifier, {
-              isSVG,
-              isCE,
-              tagName,
-              dynamic: true,
-              prevId: prev
-            })
-          )
-        )
-      );
-    } else {
-      const prev = key.startsWith("style:") ? identifier : undefined;
-      statements.push(
-        t.expressionStatement(
-          t.logicalExpression(
-            "&&",
-            t.binaryExpression("!==", identifier, t.memberExpression(prevId, identifier)),
-            setAttr(
-              path,
-              elem,
-              key,
-              t.assignmentExpression("=", t.memberExpression(prevId, identifier), identifier),
-              { isSVG, isCE, tagName, dynamic: true, prevId: prev }
-            )
-          )
-        )
-      );
-    }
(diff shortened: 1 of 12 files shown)


packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
```diff
@@ -129,65 +129,55 @@ function wrapDynamics(path, dynamics) {
...
-  const declarations = [],
-    statements = [],
-    identifiers = [],
-    prevId = t.identifier("_p$");
+
+  const statements = [];
+
   dynamics.forEach(({ elem, key, value, isSVG, isCE, tagName }) => {
     const identifier = path.scope.generateUidIdentifier("v$");
+    path.scope.push({ kind: "let", id: identifier });
```
intrnl, line 137: There's one issue with this, it adds these dirtychecking variables via `path.scope.push`, which means that the variables are placed *outside* of the template IIFE.

`results.declarations` is used to add constant variables for traversing the elements, but they're hardwired for that, and I'm not exactly sure how deep the rabbit hole goes for adding a new `dyn_declarations` array that's just meant to hold these variables.

Perhaps the ideal approach here would probably be to replace itself with an IIFE when entering a JSX element node from a foreign node, I think this would allow Babel to understand that it should probably place these variables *inside* the IIFE we just created, and we wouldn't have the need for `results.declarations` and `results.decl` at all.

packages/babel-plugin-jsx-dom-expressions/src/universal/template.js
```diff
@@ -55,43 +55,32 @@ function wrapDynamics(path, dynamics) {
...
-          t.assignmentExpression("=", t.memberExpression(prevId, identifier), setAttr(
-            path,
-            elem,
-            key,
-            identifier,
-            { dynamic: true, prevId: prev }
-          ))
-        )
-      )
+    path.scope.push({ kind: "let", id: identifier });
```
intrnl, line 63: The same goes for universal renderer.

packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/attributeExpressions/output.js
```diff
@@ -224,25 +221,20 @@ const template20 = (() => {
...
-  );
+    if (_v$5 !== (_v$5 = max())) {
+      _$setAttribute(_el$27, "max", _v$5);
+    }
+    if (_v$6 !== (_v$6 = min())) {
+      _$setAttribute(_el$28, "min", _v$6);
+    }
+    if (_v$7 !== (_v$7 = max())) {
+      _$setAttribute(_el$28, "max", _v$7);
+    }
```
intrnl, line 236: In a separate pull request I'd like to merge these 4 conditionals down into 2 conditionals, not exactly sure how that'd go though. I suppose it could be prone to side effects?

packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
```diff
@@ -129,65 +129,55 @@ function wrapDynamics(path, dynamics) {
...
-          )
-        )
-      );
-    } else {
-      const prev = key.startsWith("style:") ? identifier : undefined;
-      statements.push(
+
+    const statement = t.ifStatement(
+      t.binaryExpression("!==", identifier, t.assignmentExpression("=", identifier, value)),
+      t.blockStatement([
```
intrnl, line 164: I've used an if statement with a block statement here to make it easier to reason with the resulting codegen if you happen to be inspecting through them, given that size comparisons are done with a minifier this shouldn't really pose much of a problem.

packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
intrnl: though I suppose one question remains and that's how exactly it should be dealing with the `template` calls that are supposed to be hoisted out of our IIFE, but that shouldn't be that big of an issue

ryansolid:
This is interesting. I want to put something out there so you know what we've been thinking about with Solid 2.0's reactivity.

We want to change the renderEffects to split the pure from the side effectful. Ie..

```jsx
createRenderEffect(() => /*pull out values, tracking*/, (values, prev) => /*non-tracking, diff & apply changes*/)
```
The reason is so we can run the front half as many times as we need to (considering things like async, Suspense, Transitions) and then when everything is good to render we run the second half with the most recent values, and the prev from when the second half last ran. 

This still needs to be benchmarked, proven but we were thinking we'd compile existing effects into this form. One benefit is that we wouldn't need to return prev, as we'd have it. We'd still be generating a bit more code as we'd be creating the array/object that gets passed to the second function.

Anyway.. just wanted to put that out there.

packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/SVG/output.js
```diff
@@ -4,6 +4,7 @@ import { createComponent as _$createComponent } from "r-dom";
 import { spread as _$spread } from "r-dom";
 import { setAttribute as _$setAttribute } from "r-dom";
 import { effect as _$effect } from "r-dom";
+let _v$, _v$2, _v$3, _v$4, _v$5;
```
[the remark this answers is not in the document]
intrnl: yeah, the template isn't written in a function component here so they're hoisted to the module scope.

I'm still looking towards keeping them contained to the IIFE but it'd depend on there being any interest on this PR

intrnl:
Closing in favor of #296

Issue #283, opened 2023-11-02
Don't consider namespace imports to be dynamic expression


intrnl:
This is a common pattern with CSS Modules:
```jsx
import * as styles from './style.module.css';

const template = (
  <button class={styles.button}>
    Click here
  </button>
);
```

This gets transformed into something along the lines of
```js
_$effect(() => _$className(_el, styles.button));
```

While it is inevitable that the `className` helper has to exist, ideally, `effect` shouldn't. Property access on namespaced imports are not expected to cause reactive side-effects, and should be treated as if they were individual import specifiers instead.

intrnl:
Need to add test case for `styles[foo()]`

lxsmnsyc:
This is actually a good idea.

intrnl:
As much as I would want to get that working, I'm not sure if that's possible at the moment, the transformation code we have right now isn't designed to handle dynamic expressions within static templates, say, this isn't possible:

```js
const _tmpl$ = /*#__PURE__*/_$template(`<button${_$staticAttr('class', styles.button)}>Click here`);
```

PR #284, opened 2023-11-02
feat: don't consider ns import as dynamic

intrnl:
Fixes https://github.com/ryansolid/dom-expressions/issues/283

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
@@ -101,10 +101,20 @@ export function isDynamic(path, { checkMember, checkTags, checkCallExpressions =
   }
 
   if (checkMember && t.isMemberExpression(expr)) {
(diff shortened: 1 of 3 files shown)


intrnl:
Need to add a test case for `styles[foo()]`

packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
```diff
@@ -92,16 +92,50 @@ export function isDynamic(path, { checkMember, checkTags, checkCallExpressions =
...
+
+    if (
+      t.isIdentifier(object) &&
+      (!expr.computed ||
+        !isDynamic(path.get("property"), {
+          checkMember,
+          checkTags,
+          checkCallExpressions,
+          native,
+        }))
```
intrnl, line 115: Let me know if this code specifically needs any adjustments

ryansolid:
Interesting.. since we know it is import * as symbol it can't be reactive. Cool. Thanks

PR #287, opened 2023-11-03
Emit var for generated element references

gtm-nayan:


ryansolid:
Not sure why this one stayed in draft but implemented it now.

gtm-nayan:
intrnl had pointed out that using var inflates the gzipped size, because it's less common in the rest of the code, I was waiting till I found a way to improve that through babel

PR #288, opened 2023-11-09
Add inert as a boolean attribute

### LINKED ISSUE #281: Treat `inert` as a boolean attribute
I found this little issue while using SolidJS. I hope this is the right package to report the bug on.

[`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) should behave like `disabled`, `hidden`, and other boolean attributes, but it's rendered as `inert="true"` or `inert="false"` (which makes the tag inert).

[SolidJS Playground](https://playground.solidjs.com/anonymous/4d7e4474-8478-404c-b870-7e9adc60c7ce).

Would it be enough to add `"inert"` to this list? If so, I can open a PR 😄

https://github.com/ryansolid/dom-expressions/blob/51b297d6ed45d1a3dbbe7a45f2f58b8b524427ab/packages/dom-expressions/src/constants.js#L1-L26

intrnl:
Hit this snag while developing a custom modal component involving inert attribute

Closes #281

### DIFF
--- a/packages/dom-expressions/src/constants.js
+++ b/packages/dom-expressions/src/constants.js
@@ -10,6 +10,7 @@ const booleans = [
   "formnovalidate",
   "hidden",
   "indeterminate",
+  "inert",
   "ismap",
   "loop",
   "multiple",


ryansolid:
Thanks

PR #291, opened 2023-11-15
feat: JSX serialization

lxsmnsyc:
This adds JSX serialization capability. The change required the introduction of the `SSRNode` class which intends to replace the old `{ t: string }` so that `seroval` understands the value more.

TODO: this PR is still missing the client-side process.


lxsmnsyc:
Tricky scenario: what happens if `<Comp />` returns anything but `SSRNode`?

lxsmnsyc:
Removed the JSX serialization for now, we currently emit `undefined` for JSX.

ryansolid:
Superseded by the seroval / frames serialization work on the 2.x line. Not taking JSX serialization onto Solid 1.x, and this PR was never finished (no client half).

PR #301, opened 2023-12-29
feat(ssr): delay sending endtags to right before closing stream

NathanHuisman:
I was trying to add a `<script type="application/jd+json">` to my page, with the `onCompleteAll` callback, but when streaming, everything that gets written after the shell appears after the </body></html> closing tags. It does work, but according to the HTML spec it does result in a parse error[^after-after-body].

This is typically not a problem with browsers, they handle this gracefully, but it may be a problem with search engine crawlers. User agents may just abort the parser. This PR delays sending everything after the `</body>` end tag until right before it closes the stream.

We could also make it a bit more smart and remove the `</body>` and `</html>` tags instead and place them at the end. If someone then happens to render some data after the body or html end tag, it would show up in the body, just like the browser would parse it anyways.

[^after-after-body]: https://html.spec.whatwg.org/#the-after-after-body-insertion-mode

### DIFF
--- a/packages/dom-expressions/src/server.js
+++ b/packages/dom-expressions/src/server.js
@@ -62,6 +62,7 @@ export function renderToStringAsync(code, options = {}) {
 export function renderToStream(code, options = {}) {
   let { nonce, onCompleteShell, onCompleteAll, renderId, noScripts } = options;
   let dispose;
+  let closingTags;
   const blockingPromises = [];
   const pushTask = task => {
     if (noScripts) return;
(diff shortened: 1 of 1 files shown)


ryansolid:
Yeah I probably hadn't hit this because with async/or defer stream all the resources I think it ends up in the right order. Which I'd use in no-JS or crawler scenarios. And basically leveraging the self healing aspect of HTML has been the trick streaming in general I've never thought about it much. This is probably worth considering.

NathanHuisman:
Closing this as solid now allows per-request decision to do either async or stream rendering, which can be used to async render for bots/crawlers, which satisfies my use case

PR #302, opened 2024-01-02
chore: fix incorrect usage

Gehbt:
The uppercase 'JSX' is type


ryansolid:
Thanks.. when I started the plugin I saw a lot more examples with capital JSX. Probably a pretypescript thing. Thanks for taking the time to update this.

PR #306, opened 2024-01-15
Friendly message for when mount point doesn't exists

titoBouzout:
When the element passed to `render` doesn't exist, it shows a cryptic `element is null`. Improve that with a friendly message. 
I'm not sure what `sharedConfig.context` do, followed it is used in similar places.


ryansolid:
Thanks

PR #307, opened 2024-01-15
Fix insertExpression throw

danieltroger:
Mends part one of https://github.com/solidjs/solid/issues/2030 by selectively walking around that specific case

Tested and actually fixes the issue. Working test reproduction with yarn patch here:
`yarn && yarn dev`

[bug-test.zip](https://github.com/ryansolid/dom-expressions/files/13937759/bug-test.zip)

### DIFF
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -402,7 +402,16 @@ function insertExpression(parent, value, current, marker, unwrapArray) {
   if (value === current) return current;
   const t = typeof value,
     multi = marker !== undefined;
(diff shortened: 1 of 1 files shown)


danieltroger:
@ryansolid did you have time to take a look here? :)

ryansolid:
Understanding this, this is guarding specifically around trying to insert the parent in itself. Like the check is one level deep. I suppose that is all we need, because it is fragments that are breaking this.

ryansolid:
I think the work in #382 supercedes this specific fix. Plus I'm tenatively working with this statement gone in the next significant version. In so since we aren't merging this I'm going to close this put our efforts there.

PR #312, opened 2024-02-14
Add `useMeta`

lxsmnsyc:
Follow up to #311


ryansolid:
It looks like we will be going with a different approach. Thank you for your investigation and prototyping.

PR #313, opened 2024-03-16
Fix: hydration mismatch of nestedDynamic

birkskyum:
Closes https://github.com/solidjs/solid/issues/2100#issuecomment-2002474213

> [!WARNING]
> 
> Potential for side effects
> Unfinished as some dynamic tests fail

I made two small adjustments to utils.js and template.js, that will do as I outlined in the issue ticket linked above

For each fragment test `code.js` file I added:

```ts
const nestedDynamic = (
  <>{state.prop ? state.prop ? <div>First</div> : <></> : <></>}</>
);
```

And then i updated the snapshots by deleting them all (`find . -name \*output.js -type f  -delete`) and running the tests.

It's still a bit unclear if the altered tests that were naïvely updated (i.e. [here](https://github.com/ryansolid/dom-expressions/pull/313/files#diff-2c38ea69de6786fce4f79a6cac40e61fb2d1af3a67b42db6de08745192fe6221)) were changed for the better or worse. I'm manually tested some of these test cases that changed, and the ones I've looked at still appear to work, so it might not be all bad.

> Note: Some of the output.js files behaved strange for me in that it'll keep receiving different values depending on what it's expecting, so it's impossible to update the test to make it pass even by deleting the snapshot. I.e. this test below, if the expected value is set to the received, then it should pass, but suddenly a new value is received:

```ts

// __dom_fixtures__/fragments/output.js

-- const lastDynamic = [_tmpl$3(), _$memo(() => inserted)];
++ const lastDynamic = [_tmpl$3(), _$memo(inserted)];
```


birkskyum:
I made it a draft because we might as well introduce the fix here

birkskyum:
I've attempted to make a fix for this, but my changes unfortunately made other tests fail, and I don't know how to fix those so I'll need some assistance for a patch.

packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/conditionalExpressions/output.js
```diff
@@ -52,13 +52,12 @@ const template7 = (() => {
...
     (() => {
       var _c$3 = _$memo(() => state.count > 5);
-      return () =>
-        _c$3()
-          ? (() => {
-              var _c$4 = _$memo(() => !!state.dynamic);
-              return () => (_c$4() ? best : good());
-            })()
-          : bad;
+      return _c$3()
```
birkskyum, line 55: I tested this component (template7 in conditionalExpressions, and despite a changed output, it still render how i would expect it to.

birkskyum: Actually no, it gives a hydration error - so that's an unwanted side-effect.

birkskyum: Turns out that hydration error didn't come from this change but was there all-ready ([StackBlitz](https://stackblitz.com/edit/github-7dcnxd-sgpkre?file=src%2Fapp.tsx))

ryansolid:
The biggest risk is things that might work, but won't update. There is also the chance of weird recursion due to passing functions up the chain.

ryansolid:
I think this was just a bug. https://github.com/ryansolid/dom-expressions/commit/eeca96004070473aed0e3454423ce164f0a5e7d2#diff-6971a60935cf3ec66903ea31c0bb58c4667e48f29a9e71aacdc2a4ba51745295R133

We have 2 different modes for compilation here. One that leverages what I call onioning which is slightly more performant as it can avoid remaking memos and one that inlines it. We were using inline for components and we should have been for fragments as well.

birkskyum:
Excellent! That's a better output:

```ts
// Before
  return _$memo2((() => {
    const _c$ = _$memo2(() => !!obj2.prop);
    return () => _c$() ? (() => {
      const _c$2 = _$memo2(() => !!obj2.prop);
      return () => _c$2() ? _tmpl$() : [];
    })() : [];
  })());

// After
  return _$memo(() =>
    _$memo(() => !!obj1.prop)() ?
    (_$memo(() => !!obj2.prop)() ?
      _$getNextElement(_tmpl$2) : []) : []
  );
```

Think maybe the pnpm lock file got out of sync with the new releases

PR #314, opened 2024-03-19
sync lockfile

birkskyum:


birkskyum:
it's just dev deps - doesn't matter

ryansolid:
Yeah lerna always does this after a release and I also do that manual fix lock commit. Forgot this time.

PR #316, opened 2024-04-18
fix(dom-expressions): proper ref handling in spread

mdynnl:
fixes https://github.com/solidjs/solid/issues/2136

i'm not sure if `untrack(() => props.ref(node))` is better

### DIFF
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -183,7 +183,7 @@ export function spread(node, props = {}, isSVG, skipChildren) {
   if (!skipChildren) {
     effect(() => (prevProps.children = insertExpression(node, props.children, prevProps.children)));
   }
-  effect(() => props.ref && props.ref(node));
+  effect(() => (typeof props.ref === "function" ? use(props.ref, node) : (props.ref = node)));
   effect(() => assign(node, props, isSVG, true, prevProps, true));
   return prevProps;
 }


ryansolid:
Yeah untrack is a good question. We weren't before so I think stick with this. Thanks.

danieltroger:
Sorry I don't understand what this part of the expression does? `(props.ref = node)`

Because now I'm running into this error (did yesterday too):
Screenshot 2024-04-19 at 18 38 26


Basically what happens is that I pass the props to a component and then this code `props.ref_ = r$` mutates my prop object and makes ref an element inside it. And later on when I expect it to be a function it throws.

This is the function un-transpiled that I'm having issues with:

```
// Work around https://github.com/solidjs/solid/issues/2136 and random shit I thought was an issue before I found https://github.com/solidjs/solid/issues/2137
function InComponentOrDiv(props: {
  component_:
    | undefined
    | false
    | ((componentProps: { children: JSX.Element; ref?: (el: HTMLElement) => void; class: string }) => JSX.Element);
  children: JSX.Element;
  class_: string;
  ref_?: (el: HTMLElement) => void;
}) {
  const resolved = children(() => props.children);
  return createMemo(() => {
    const Component = props.component_;
    if (Component) {
      return (
        <Component class={props.class_} ref={props.ref_}>
          {resolved()}
        </Component>
      );
    }
    return (
      <div class={props.class_} ref={el => props?.ref_?.(el)}>
        {resolved()}
      </div>
    );
  }) as unknown as JSX.Element;
}
```

mdynnl:
@danieltroger I hope https://github.com/ryansolid/dom-expressions/pull/332 should address this.

PR #317, opened 2024-04-19
fix(dom-expressions): handle expression in ref

mdynnl:
fixes https://github.com/solidjs/solid/issues/2137

these are still relevant here so didn't touch them
```jsx
let refTarget;
const template8 = <div ref={refTarget} />;
```

```jsx
const template10 = <div ref={refFactory()} />;
```

not sure why the check for call expression was there though.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -495,7 +495,7 @@ function transformAttributes(path, results) {
                 )
               )
             );
-          } else if (t.isCallExpression(value.expression)) {
+          } else {
             const refIdentifier = path.scope.generateUidIdentifier("_ref$");
             results.exprs.unshift(
               t.variableDeclaration("var", [
--- a/packages/babel-plugin-jsx-dom-expressions/src/universal/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/universal/element.js
@@ -131,7 +131,7 @@ function transformAttributes(path, results) {
                 )
               )
             );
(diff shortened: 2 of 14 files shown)


SarguelUnda:
Posting here to not flood the issue we came from.

As a general case I understand your point. I was today year old when I realised that props?.ref isn't lhs aswell.

Isn't there an argument to transform props?.ref into props.ref to get the optimisation of the lvalcond instead of the more general use function that your merge request propose ?

SarguelUnda:
Moreover if props.ref is not à function your fix doesn't work either no ?

mdynnl:
https://github.com/ryansolid/dom-expressions/blob/ac40fb85d27d119ee940d3f555e668ebaeaea22f/packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/attributeExpressions/code.js#L65

transforms to

https://github.com/ryansolid/dom-expressions/blob/ac40fb85d27d119ee940d3f555e668ebaeaea22f/packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/attributeExpressions/output.js#L170-L175

and because of this
https://github.com/ryansolid/dom-expressions/blob/ac40fb85d27d119ee940d3f555e668ebaeaea22f/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js#L498

it only works for a call expression,

but it should work for anything that isn't a literal, let variable, member expression.

ryansolid:
Yeah I think you are right. I probably was fixing a specific issue and was looking too narrow. That being said the fact this change hasn't changed the output of any tests suggests to me are definitely missing some tests we should have.

ryansolid:
Thank you for adding the tests. Much appreciated.

PR #318, opened 2024-04-23
fix `this` compilation inside class methods

mdynnl:
related https://github.com/ryansolid/dom-expressions/pull/295
fixes https://github.com/solidjs/solid/issues/2141

Not adding tests to other modes as it's essentially the same.


ryansolid:
Thanks

PR #319, opened 2024-04-25
Prefer to set property when property & attribute are identical

qeleb:
## Changes
1. It saves 5+ bytes for every single attribute that's instead set as a property, so I've added a list of [reflected attributes](https://html.spec.whatwg.org/#reflected-idl-attribute) to the Properties list.
1. Fix 1 typo in a comment

#### Example code
```tsx
render(
  () => (
    <>
      <a id={Math.random() > 0.5 ? 'linkA' : 'linkB'}>link</a>                   {/* id */}
      <button name={Math.random() > 0.5 ? 'name a' : 'name b'}>name btn</button> {/* name */}
      <link rel={Math.random() > 0.5 ? 'preload' : 'preconnect'}>link</link>     {/* rel */}
      <img src={logo} alt={Math.random() > 0.5 ? 'a img' : 'b img'} />           {/* src */}
      <input type={Math.random() > 0.5 ? 'text' : 'password'} />                 {/* type */}
    </>
  ),
  document.body
);
```
#### OLD output
```js
(() => {
  return [
    ((c = G()), e(() => a(c, 'id', Math.random() > 0.5 ? 'linkA' : 'linkB')), c),
    ((r = k()), e(() => a(r, 'name', Math.random() > 0.5 ? 'name a' : 'name b')), r),
    ((o = z()), e(() => a(o, 'rel', Math.random() > 0.5 ? 'preload' : 'preconnect')), o),
    ((n = E()),
    a(n, 'src', 'data:image/svg+xml,somesvg'),
    e(() => a(n, 'alt', Math.random() > 0.5 ? 'a img' : 'b img')),
    n),
    ((t = B()), e(() => a(t, 'type', Math.random() > 0.5 ? 'text' : 'password')), t),
  ];
}, document.body);
```
#### NEW output `(-0.1KB)`
```js
(() => {
  return [
    ((c = A()), e(() => (c.id = Math.random() > 0.5 ? 'linkA' : 'linkB')), c),
    ((r = G()), e(() => (r.name = Math.random() > 0.5 ? 'name a' : 'name b')), r),
    ((o = k()), e(() => (o.rel = Math.random() > 0.5 ? 'preload' : 'preconnect')), o),
    ((n = O()),
    (n.src = 'data:image/svg+xml,somesvg'),
    e(() => (n.alt = Math.random() > 0.5 ? 'a img' : 'b img')),
    n),
    ((t = E()), e(() => (t.type = Math.random() > 0.5 ? 'text' : 'password')), t),
  ];
  var t, n, l, o, r, c, f;
}, document.body);
```

## Notes
1. There are more reflected properties to be added in the future for further gains, although some properties like `maxlength (string)`/`maxLength (number)` will require slightly more effort to coerce types

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/ssr/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/ssr/element.js
@@ -227,7 +227,7 @@ function normalizeAttributes(path) {
         a.node.name.name === "className" ||
         a.node.name.name === "classList")
   );
-  // combine class propertoes
+  // combine class properties
   if (classAttributes.length > 1) {
     const first = classAttributes[0].node,
       values = [],
--- a/packages/dom-expressions/src/constants.js
+++ b/packages/dom-expressions/src/constants.js
@@ -46,7 +46,6 @@ const BooleanAttributes = /*#__PURE__*/ new Set(booleans);
 
 const Properties = /*#__PURE__*/ new Set([
   "className",
-  "src",
   "value",
   "readOnly",
   "formNoValidate",
--- a/packages/dom-expressions/src/constants.js
+++ b/packages/dom-expressions/src/constants.js
@@ -1,3 +1,19 @@
+const reflectedAttributes = [
+  "align",
+  "alt",
+  "href",
+  "id",
+  "loading",
+  "name",
+  "rel",
+  "role",
+  "src",
+  "srcset",
+  "sizes",
+  "title",
+  "type"
+];
+
 const booleans = [
   "allowfullscreen",
   "async",
@@ -30,12 +46,14 @@ const BooleanAttributes = /*#__PURE__*/ new Set(booleans);
 
 const Properties = /*#__PURE__*/ new Set([
   "className",
+  "src",
   "value",
   "readOnly",
   "formNoValidate",
   "isMap",
   "noModule",
   "playsInline",
+  ...reflectedAttributes,
   ...booleans
 ]);
 
@@ -75,6 +93,10 @@ const PropAliases = /*#__PURE__*/ Object.assign(Object.create(null), {
     $: "readOnly",
     INPUT: 1,
     TEXTAREA: 1
+  },
+  src: {
+    $: "src",
+    IMG: 1
   }
 });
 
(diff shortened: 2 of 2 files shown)


ryansolid:
Trying to think if there is a downside. When I started Solid I actually used all properties but we switched to attributes for consistency reasons. We still use properties for webcomponents. So trying to remember why we made this choice. I might have to do some digging to remind myself of why.

qeleb:
> Trying to think if there is a downside. When I started Solid I actually used all properties but we switched to attributes for consistency reasons. We still use properties for webcomponents. So trying to remember why we made this choice. I might have to do some digging to remind myself of why.

Ah please do.
I’ve been writing
`prop:src=`, `prop:alt=`, & `prop:href=` for months at work and I was hoping this is something the compiler could take on specifically for the cases where the attribute & property have the exact same behavior.

We definitely cannot do this for all of them though— there are some tricky ones like checked where the `attr:checked` === `prop:defaultChecked` (but not `prop:checked`)

there is a similarly weird case for the value property.

To the best of my understanding, this optimization is safe for all IDL attribute reflection properties so I made a very careful list of only a few.

qeleb:
@ryansolid Have you gotten a chance to review this?

If we do move forward with this, the next step is to have the compiler “upgrade” uses of IDL reflection attributes to their property counterpart (ie. `attr:max-length` => `prop:maxLength`)

qeleb:
@ryansolid Would you be willing to merge this optimization if it were opt-in only like `omitNestedClosingTags`?

ryansolid:
Ok looking at this, the list is pretty small and `href` was one of the ones I remember being problematic. That being said the extra size of the array of strings versus the character savings might be a bit of a wash on some projects. I think I'm fine with this.

ryansolid:
Just updating the tests. One downside I actually see is now we need to handle `null` and `undefined` for string attributes with a special helper or extra compiled code. So I actually don't think this improves code size. The whole reason for the setAttribute helper is to basically handle null/undefined. I think I'm going to shelf this ultimately. We can revisit if there are other benefits.

qeleb:
@ryansolid Are you sure? This has improved code size on 6 of my projects now. Also, we could, do this optimization only where not null/undefined/empty to avoid the cases where there would be extra code.

Issue #321, opened 2024-04-26
bug: Compilation from JSX style attribute to ternary results in functions getting called twice


NathanHuisman:
## Reproducible example

[Playground link](https://playground.solidjs.com/anonymous/5da2e13c-9602-4615-a1b3-f1f19da308fa)

```tsx
function Example() {
  const random = () => Math.random() > 0.5 ? null : "123";
  return (
    <div
      style={{
        "--blah": random()
      }}
    />
  );
}
```

Compiles to the following:
```js
function Example() {
  const random = () => Math.random() > 0.5 ? null : "123";
  return (() => {
    var _el$2 = _tmpl$2();
    _$effect(() => random() != null ? _el$2.style.setProperty("--blah", random()) : _el$2.style.removeProperty("--blah"));
    return _el$2;
  })();
}
```

The function `random` gets called twice, even though it's not pure.
This is not just problematic for this contrived example, but also for functions that are not side-effect free.

Even more problematic is that this behaves differently when more properties are involved:

```tsx
function Example() {
  const random = () => Math.random() > 0.5 ? null : "123";
  const someOtherThing = () => "abc";
  return (
    <div
      style={{
        "--blah": random(),
        "--someotherthing": someOtherThing()
      }}
    />
  );
}
```

Compiles to:
```js
function Example() {
  const random = () => Math.random() > 0.5 ? null : "123";
  const someOtherThing = () => "abc";
  return (() => {
    var _el$2 = _tmpl$2();
    _$effect(_p$ => {
      var _v$ = random(),
        _v$2 = someOtherThing();
      _v$ !== _p$.e && ((_p$.e = _v$) != null ? _el$2.style.setProperty("--blah", _v$) : _el$2.style.removeProperty("--blah"));
      _v$2 !== _p$.t && ((_p$.t = _v$2) != null ? _el$2.style.setProperty("--someotherthing", _v$2) : _el$2.style.removeProperty("--someotherthing"));
      return _p$;
    }, {
      e: undefined,
      t: undefined
    });
    return _el$2;
  })();
}
```

Now the functions do get called only once.

## Suggested fix

My suggestion is to compile to something like the following instead:

```js
function Example2() {
  const random = () => Math.random() > 0.5 ? null : "123";
  return (() => {
    var _el$ = _tmpl$();
    _$effect(() => {
      var _$v = random();
      _$v != null ? _el$.style.setProperty("--blah", _$v) : _el$.style.removeProperty("--blah")
    });
[...]

PR #323, opened 2024-05-08
fix double invocation for single dynamic style

mdynnl:
fixes https://github.com/ryansolid/dom-expressions/issues/321


ryansolid:
Thank you

PR #327, opened 2024-06-04
Don't unwrap IIFE if it contains parameters

intrnl:
- Adds test for an IIFE with a parameter
- Adjust the transformation so unwrapping only happens when there isn't any


ryansolid:
Thanks

PR #336, opened 2024-08-04
make optional attributes accept undefined when exactOptionalPropertyTypes is enabled

Huliiiiii:
close solidjs/solid#1640


ryansolid:
Thank you... I'm sure this was tedious to do. I really appreciate it.

PR #338, opened 2024-08-14
Create text node for hypertext element of type bigint or symbol

wrnrlr:


wrnrlr:
I guess you could also use `Symbol.keyFor(...)` instead of `.toString()` for the symbol type.

ryansolid:
Thanks

PR #342, opened 2024-08-20
update event types

trusktr:


packages/dom-expressions/src/jsx.d.ts
```diff
@@ -270,23 +270,16 @@ export namespace JSX {
...
     onPointerUp?: EventHandlerUnion<T, PointerEvent> | undefined;
-    onProgress?: EventHandlerUnion<T, Event> | undefined;
+    onProgress?: EventHandlerUnion<T, ProgressEvent> | undefined;
     onRateChange?: EventHandlerUnion<T, Event> | undefined;
     onReset?: EventHandlerUnion<T, Event> | undefined;
     onScroll?: EventHandlerUnion<T, Event> | undefined;
     onScrollEnd?: EventHandlerUnion<T, Event> | undefined;
     onSeeked?: EventHandlerUnion<T, Event> | undefined;
     onSeeking?: EventHandlerUnion<T, Event> | undefined;
-    onSelect?: EventHandlerUnion<T, UIEvent> | undefined;
```
trusktr, line 280: I took a look at https://w3c.github.io/uievents/#events-uievent-types which says that some events can be either `UIEvent` objects or `Event` objects, however I can't find any example or information on when an even is actually a `UIEvent`.

This other spec shows the events as `Event` only: https://html.spec.whatwg.org/multipage/indices.html#events-2

In practice, I always see `Event` for the events listed as possible `UIEvent`s.

Here's a `select` event example that shows the type is `Event` and not `UIEvent`:
https://codepen.io/trusktr/pen/KKjQEBw/1278088be6681bfa7cfa3d449e87180e?editors=1010

For most of these events, even MDN lists them as `Event` only, for example the `select` event here:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select_event

The `error` event is an odd ball in the MDN docs, where it *does* mention the possibility of it being a `UIEvent`:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/error_event

However in practice, I only see `error` events as `Event`, not `UIEvent`. Example:

https://codepen.io/pen/?editors=0011

So I think it is safe to just leave any events that can be a `UIEvent` as `Event`, and if the user happens to know when it can be a `UIEvent`, they can check for it conditionally.

If we leave it as `UIEvent`, then it is incorrect when it receives an `Event`.

PR #345, opened 2024-09-01
export client-side APIs from dom-expressions server.js, make them throw an error

trusktr:
Fixes:

- https://github.com/solidjs/solid/issues/1984
- https://github.com/solidjs/solid-start/issues/1618

Are there any implications that I have overlooked? If no one is currently importing `Aliases`/etc on the server (because they currently can't) then adding the exports should be harmless.

This PR goes along with

- https://github.com/solidjs/solid/pull/2269 (merge this one first, then that one after updating dom-expressions in that one)


trusktr:
Still WIP, f.e. after this we get this downstream:

```
SyntaxError: The requested module 'solid-js/web' does not provide an export named 'classList'
```

trusktr:
Alrighty, updated to export client-side functions from server.js; they throw if called on the server side.

I tested in the solid repo with this version of dom-expressions, by running `import('./packages/solid/html/dist/html.js').then(console.log)` in the `node` repl which shows that `html` can be imported. I also verified that the errors are thrown for client-only APIs.

Here's what the `node` repl output looks like in a separate folder with `solid-js` linked into `node_modules` with the updated `dom-expressions`:

```
❯ node
Welcome to Node.js v20.6.1.
Type ".help" for more information.
> import('solid-js/html').then(console.log)
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 374,
  [Symbol(trigger_async_id_symbol)]: 366
}
> [Module: null prototype] { default: [Function: html] }
> import('solid-js/web').then(m => m.effect(() => {}))
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 711,
  [Symbol(trigger_async_id_symbol)]: 703
}
> Uncaught:
Error: Called API is not supported on server side. Run client-only code in onMount, or run client-only component with <Show>.
    at Module.notSup (file:///Users/trusktr/src/solidjs+solid/packages/solid/web/dist/server.js:654:9)
```

packages/dom-expressions/src/client.d.ts
```diff
@@ -57,7 +57,7 @@ export function dynamicProperty(props: unknown, key: string): unknown;
 export function hydrate(
   fn: () => JSX.Element,
   node: MountableElement,
-  options?: { renderId?: string, owner?: unknown }
+  options?: { renderId?: string; owner?: unknown }
```
trusktr, line 60: prettier did it

packages/dom-expressions/src/client.js
```diff
@@ -466,7 +466,7 @@ function insertExpression(parent, value, current, marker, unwrapArray) {
       if (marker === undefined) return (current = [...parent.childNodes]);
       let node = array[0];
       if (node.parentNode !== parent) return current;
-      const nodes = [node]
+      const nodes = [node];
```
trusktr, line 469: prettier did it

trusktr:
Converted back to draft, doing some testing to ensure that the build is working dependong on Solid Start `ssr` mode or not...

packages/dom-expressions/src/server.js
```diff
@@ -1,7 +1,26 @@
...
-export { createComponent } from "rxcore";
+
+export {
+  getOwner,
+  createComponent,
+  createRenderEffect as effect,
+  createMemo as memo,
+  untrack,
+  mergeProps
+} from "rxcore";
```
trusktr, line 12: Is this accurate? I based this off of web/src/core.ts (which I assume indicates which APIs are available server-side, minus any that were already included here. This also matches with what client.js exports.

packages/dom-expressions/src/server.js
```diff
@@ -674,3 +693,33 @@ export function ssrSpread(props, isSVG, skipChildren) {
...
+  notSup as template,
+  notSup as setProperty,
+  notSup as className,
+  notSup as assign,
+  notSup as hydrate,
+  notSup as getNextElement,
+  notSup as getNextMatch,
+  notSup as getNextMarker,
+  notSup as runHydrationEvents
+};
```
trusktr, line 719: And are these accurate? Wondering if I missed an export from somewhere else that does in fact work server-side and should not be mapped to `notSup`.

packages/dom-expressions/src/server.js
```diff
@@ -1,7 +1,19 @@
 import { Aliases, BooleanAttributes, ChildProperties } from "./constants";
 import { sharedConfig, root } from "rxcore";
 import { createSerializer, getLocalHeaderScript } from "./serializer";
-export { createComponent } from "rxcore";
+
+export { getOwner, createComponent, effect, memo, untrack } from "rxcore";
```
trusktr, line 5: Is this list accurate? I based this off of `web/src/core.ts`, which I assume indicates which APIs are available server-side, but I did not include those that were already included here in `server.js`. This brings server.js exports to closer parity with client.js exports.

trusktr: Ok, no, not accurate. The code is difficult to follow.

I see that the build output has code injected, that I need to avoid clashing with:

```js
const isServer = true;
const isDev = false;
function render() {}
function hydrate() {}
function insert() {}
function spread() {}
function addEventListener() {}
function delegateEvents() {}
function Dynamic(props) {
  const [p, others] = splitProps(props, ["component"]);
  const comp = p.component,
    t = typeof comp;
  if (comp) {
    if (t === "function") return comp(others);else if (t === "string") {
      return ssrElement(comp, others, undefined, true);
    }
  }
}
function Portal(props) {
  return "";
}
```

trusktr: Where is this list of exports? For example `function render() {}` is injected, and that is not exported in `server.d.ts`.

Is `babel-plugin-transform-rename-import` doing the injection?

trusktr: it is confusing because it looks like the exports from `dom-expressions` get merged with the exports from `web/src/index.ts` into a single file. It bespoke and totally non-obvious how the code comes together.

packages/dom-expressions/src/server.js
```diff
@@ -1,7 +1,19 @@
-import { Aliases, BooleanAttributes, ChildProperties } from "./constants";
+import { Aliases, BooleanAttributes, ChildProperties, Properties } from "./constants";
```
trusktr, line 1: This import was missing, see below.

trusktr:
Ok I think this is finally ready again.

packages/dom-expressions/src/server.js
trusktr: These are all new functions never exported before for server, so it is non-breaking to make them errors I believe. However the question is, do you want any of them to be no-ops instead of error-throwing? I think errors are better for DX to inform people what not to do.

Also, any of these can be overriden to be no-ops over in solid-js, for example like is done here:

https://github.com/solidjs/solid/pull/2269/files#r1740197787

ryansolid:
Yeah just eyeballing this looks pretty good.

PR #347, opened 2024-09-05
preliminary support for `bool:` [needs help]

titoBouzout:
The idea of the `bool:` namespace is to control if an attribute is present or not by evaluating its value to truthy/falsy. If it is truthy the attribute is added with an empty string as value, and if falsy the attribute is removed.  It's similar to `BooleanAttributes` that dom-expressions already considers but controllable by the user. This is specially relevant in the web components world where names are arbitrary and cannot be known beforehand.

The namespace will make the intention clear and this cannot be done with dom-expressions `setAttribute`.

I am running into an issue trying to implement this feature and your help would be welcome

I have added 1 test for "custom elements" 
`babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/customElements/code.js`

When I run `npm run test` it displays two failing tests:
 - FAIL  test/dynamic-dom.spec.js
 - FAIL  test/dom.spec.js

It looks like It's reusing the test I wrote in two different places and expecting for the same test two different outputs, I can't figure out what's going on with this test. Any ideas?

### DIFF
--- a/packages/dom-expressions/src/server.js
+++ b/packages/dom-expressions/src/server.js
@@ -667,7 +667,19 @@ export function ssrSpread(props, isSVG, skipChildren) {
     ) {
       continue;
     } else {
+      // bool:
+      if (prop.slice(0, 5) === "bool:") {
+
+       if(!value) continue
+       prop = prop.slice(5);
+       result += `${escape(prop)}`;
+
+       continue
+      }
+
+      // attr:
       if (prop.slice(0, 5) === "attr:") prop = prop.slice(5);
+
       result += `${Aliases[prop] || escape(prop)}="${escape(value, true)}"`;
     }
     if (i !== keys.length - 1) result += " ";
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -705,6 +705,42 @@ function transformAttributes(path, results) {
             tagName
           });
         } else if(key.slice(0, 5) === 'bool:'){
+
+            // inline it on the template when possible
+            let content = value;
+
+            if (t.isJSXExpressionContainer(content)) content = content.expression;
+
+            function addBoolAttribute() {
+              results.template += `${needsSpacing ? " " : ""}${key.slice(5)}`;
+              needsSpacing = true;
+            }
+
+            switch (content.type) {
+              case "StringLiteral": {
+                if (content.value.length && content.value !== "0") {
+                  addBoolAttribute();
+                }
+                return;
+              }
+              case "NullLiteral": {
+                return;
+              }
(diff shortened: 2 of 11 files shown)


packages/dom-expressions/src/server.js
```diff
@@ -667,7 +667,19 @@ export function ssrSpread(props, isSVG, skipChildren) {
     ) {
       continue;
     } else {
+      // bool:
+      if (prop.slice(0, 5) === "bool:") {
+
+       if(!value) continue
+       prop = prop.slice(5);
+       result += `${escape(prop)}`;
```
titoBouzout, line 675: does this need a space?

titoBouzout:
ok, I think this can be reviewed. The test I wrote works, there's still that failing auto-generated test that I don't know from where that's coming from.

ryansolid:
If you are curious why the tests fail its because we have some modes of universal rendering which are hybrid. Ie.. they combine both custom rendering and dom renderer. The problem was that your custom element `my-el` wasn't registered as a dom-element so it assumed it was a universal element and didn't get the right treatment.

titoBouzout:
Ah, I see, I will take a look at that if I run into this problem again, thank you!

packages/dom-expressions/src/server.js
titoBouzout: @ryansolid this has been merged, pinging because Im not sure if you saw the question above, thanks

ryansolid:
@titoBouzout one question I had on further review is why set the it as an attribute instead of a prop. I guess we don't know if all booleans are reflective which is fair. I suppose if it was sufficient people could have just used `prop:booleanThing`. It is just we handle all other booleans as an properties. It's a minor thing but I'm always looking at size savings.

packages/dom-expressions/src/server.js
ryansolid: Yeah it can but it depends which is why we have that logic after the condition. I'm going to slightly refactor to include it.

titoBouzout:
@ryansolid I think the main issue this solves is getting rid of the heuristic and being able to be explicit about the intention in a way that's simple and idiomatic to solid, complementary to `attr:` and `prop:`

Considerations... WC authors will do things in different ways, ones would expect a prop and others will expect an attribute. This could be relevant to the `attributeChangedCallback`. The tricky part is to think that external web components can be used in Solid, so one needs to imagine that they haven't used `solid-element` and the like, a bunch of _solid ecosystem_ assumptions go out of the window. Giving control back empowers and I consider it important to not cause friction in adoption.

That being said, I am honestly still trying to figure out how all play together. I see a lot of benefices using WC and I'm trying to move to them, there are challenges. Joe expands on this feature in https://github.com/solidjs/solid/issues/2158 and `lit-html` has similar features to `prop/bool` (they default to attribute I think).

PR #348, opened 2024-09-06
`is` in element attributes/props counts as custom element

titoBouzout:
It adds a few checks for `is` in attributes/props to mark the element as a custom element. 

fixes https://github.com/solidjs/solid/issues/2122
still there's a need to take into account spreads


trusktr:
I think its fine to get this in without supporting spreads, https://github.com/solidjs/solid/issues/2122#issuecomment-2336503911.

ryansolid:
Thanks

PR #349, opened 2024-09-06
use `document.importNode` for when `img/iframe` use `loading="lazy"`

titoBouzout:
This attempts to reuse the same logic as `hasCustomElement` as an `isImportNode`. That currently only becomes `true` when an `image/iframe` has the attribute `loading="lazy"` 

To solve https://github.com/solidjs/solid/issues/1828 avoiding using `document.importNode` for every other element.

This in theory should have worked on the first try. For some reason nested elements with such an attribute won't be marked as `isImportNode`. Handled this oddity with a _least resort_ regexp for now.

`_$template('', true, false),`
`_$template('<div>', true, false);` 

Without the regexp the last one won't have `true, false)`

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -70,6 +70,7 @@ export function transformElement(path, info) {
       postExprs: [],
       isSVG: wrapSVG,
       hasCustomElement: isCustomElement,
+      isImportNode: tagName === 'img' && path.get("openingElement").get("attributes").some(a => a.node.name?.name === "loading" && a.node.value?.value === "lazy"),
       tagName,
       renderer: "dom",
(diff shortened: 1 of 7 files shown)


titoBouzout:
ok, this is ready for review. I was able to get rid of the regexp. 

-- 
I had the issue(now fixed) of the following nested template not having the "use importNode" bit
`_$template('<div>');`

I presume because the condition `child.id` didnt enter. So I moved the copy of the flag above it, and that worked. Now I wonder if theres some latent bug of `hasHydratableEvent` and `hasCustomElement` suffering the same problem. That child templates do not copy some properties to its parents

 
sublime_text_4Vbyo6Px85

ryansolid:
It's a bit odd. What I mean is every native element will have `child.id` I'd think. But maybe I'm wrong. Maybe only ones with dynamic bindings or something.

PR #351, opened 2024-09-22
fix: custom elements props case

titoBouzout:
Issue:  custom element `props` lose case, example `myFn` becomes `myfn`.

I think this issue is unintentional, there's this function that it's trying **to give case** to custom elements props, for when the prop name comes from a case-insensitive place, ex transforming `the-prop` into `theProp`.  The function is

```js
export function toPropertyName(name) {
  return name.toLowerCase().replace(/-([a-z])/g, (_, w) => w.toUpperCase());
}
```

I would speculate that by trying to guard the case-sensitive regexp, a `toLowerCase` came to mind and it's causing the issue. Because if the prop doesn't have the char `-`, then it won't result into the desire intention.

It is used (as far as I could tell) on custom elements, as every use of `toPropertyName` seems to have a `isCE` check guard.

Here is a repro  https://playground.solidjs.com/anonymous/bfd05199-72f0-405a-a2c6-681afae82fbd 
Inspecting the element, and then from the "properties" tab from the developer tools 

chrome_1E0Xiq9IPO

When using `prop:` namespace it works correctly 

chrome_OJhdTqvgN7

I tested the fix by manually editing a vite build of "my-solid-project"

This was reported by @chee on solid-docs https://github.com/solidjs/solid-docs/issues/783

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
@@ -228,7 +228,7 @@ export function toAttributeName(name) {
 }
 
 export function toPropertyName(name) {
-  return name.toLowerCase().replace(/-([a-z])/g, (_, w) => w.toUpperCase());
+  return name.replace(/-([a-z])/g, (_, w) => w.toUpperCase());
 }
 
 export function wrappedByText(list, startIndex) {


ryansolid:
I think this was intentional. @trusktr I am trying to remember the details and where we posted it but you had a long list of how you wanted to see custom elements treated. And if I remember correctly the idea was that we would purposefully to lowercase everything natively so that it handled like HTML in a sense. I wish I remembered if this was a discord thread or a github issue but I remember thinking it weird at the time and then was convinced. I think we wanted to encourage people to write things as if there were attributes and then do a reasonable conversion back to properties to support non-string data types.

I believe we transform `some-attr` to `someAttr` when we set the property. This does assume a decent amount to be fair but it keeps things consistent for SSR which has to treat them as attributes always anyway. So case sensitivity isn't good for us here.

trusktr:
Hmmm, HTML is case insensitive, so in HTML `fooBar=""`, `FOOBAR=""`, and `fOoBaR=""` set the same attribute:

```js
document.body.setAttribute('FoObAr', '123')
console.log(document.body.attributes.foobar.value) // 123
document.body.setAttribute('FOOBAR', '456')
console.log(document.body.attributes.foobar.value) // 456
```

## Rethinking template attributes/props mapping?

But because Solid reads the template before mapping it to DOM, it knows ahead of time what the user wrote.

I've been thinking, maybe it may be good to change the rules to be more consistent with the wider ecosystem (outside of Solid), simplify the template implementation a little bit, and give users 100% control of elements without any compromises.


some frameworks like Lit, and in the near future React 19, map template attributes (or props, depending) to JS properties without the the dash-case to camelCase conversion

- because we see the template ahead of time, we know if the user wrote `foobar` vs `FOOBAR`.
- it is possible that for sake of consistency, we could have `foobar=""` **always** set an attribute regardless of the type of element the attribute is on (custom vs builtin)
  - right now `foo=""` sets the `foo` attribute on builtin elements by default, otherwise it sets the `foo` JS property custom elements.
    - This may not be obvious to new users at first, for example it is not mentioned in the new docs [here](https://docs.solidjs.com/concepts/understanding-jsx) or [here](https://docs.solidjs.com/concepts/components/props).
  - `foobar=""` and `fOoBaR=""` would *always* set the same attribute (on custom and builtin elements)
    - `attr:foobar=""` and `attr:fOoBaR=""` would have the same behavior of always setting the same attribute
- case sensitivity would be distinguished only with property syntax where `prop:foobar=""` would set a `.foobar` JS property, and `prop:fOoBaR=""` would set a `.fOoBaR` JS property
  - `html` would have Lit-compatible syntax besides `prop:` syntax: `.foobar=""` would set the `.foobar` JS property, and `.fooBAR=""` would set the `.fooBAR` JS property.
  - `prop:foo` (in JSX ot `html`) and `.foo` (in `html` only) would *always* be needed to set a JS property regardless if custom or builtin element
- Boolean attribute syntax would be case insensitive, f.e. `bool:foobar=` and `bool:FOObar=` would handle the same attribute (but with boolean treatment where truthy means the attribute should exist, and falsy means the attribute should be removed) regardless if custom or builtin element
  - `html` would have Lit-compatible syntax besides `bool:` syntax: `?foobar=` and `?fooBAR=` handle the same attribute
  - when its a static attribute (`boo:foo=""` instead of `bool:foo=${}`), it would work like setting a string value without boolean syntax: `bool:foobar=".."` would be equivalent to `foobar=".."`, setting the attribute (case insensitive) to a string to emulate that in HTML `foobar=""` and `foobar="false"` are both "truthy" in _most_ HTML cases
    - there are other "enumerated attributes" that set JS properties to `false` when the string value is `"false"` otherwise `true`, and when omitted may either set the JS property to `true` or `false` depending on its default
      - Lume does boolean handling this way: `foo`, `foo="true"`, and `foo="blah"` all set the JS property to `true`; `foo="false"` sets the JS prop to `false`, and omitting the attribute sets the JS prop to `true` or `false` based on the JS property's default value.


## Why?

This would have some benefits:

- the template implementation would be simpler: it would no longer have any conditional logic that tries to guess whether an attribute or a JS property should be set, no special case attribute-to-property name mappings, etc.
- simpler for users to understand (or assume) out of the box without the documentation having to mention different handling between custom and builtin elements
- consistency as in other frameworks do this (or some of this), for example Lit does exactly this while React 19 doesn't offer full control (it will still have issues, [for](https://github.com/facebook/react/issues/30418), [example](https://github.com/facebook/react/issues/29037), and despite the warning signs [starting here](https://github.com/facebook/react/issues/11347#issuecomment-1732205832) that are honestly quite easy to avoid), but React 19 will map case-sensitive names to JS properties when they exist, otherwise fall back to attributes. For React 19, you will need to understand this [monster table](https://josepharhar.github.io/react-custom-elements/) to know how it works, and we will still have to break out of the template with refs to achieve things in various cases.
- consistency as in **what you see is what you get**: you know from reading the template **_exactly**_ what it is doing just from reading the template. Compare the following table to the previous bullet point's React 19 table (`_` represents a custom element or builtin element, does not make a difference):
  - `<_ foo=... />` -> always set the `foo` attribute
  - `<_ prop:foo=... />` -> always set the `foo` JS property
  - `<_ bool:foo=... />` -> always add the `foo` attribute if the interpolated value is truthy, otherwise remove the attribute. If non-interpolated, apply the string value to the attribute as if `bool:` was not used (attribute always exists in that case).
  - `<_ fooBar=... />` -> always set the `foobar` attribute
  - `<_ prop:fooBar=... />` -> always set the `fooBar` JS property
  - `<_ bool:fooBar=... />` -> always add the `foobar` attribute if the interpolated value is truthy, otherwise remove the attribute. If non-interpolated, apply the string value to the attribute as if `bool:` was not used (attribute always exists in that case).
  - `<_ foo-Bar=... />` -> always set the `foo-bar` attribute
  - `<_ prop:foo-Bar=... />` -> always set the `"foo-Bar"` JS property
  - `<_ bool:fooBar=... />` -> always add the `foo-bar` attribute if the interpolated value is truthy, otherwise remove the attribute. If non-interpolated, apply the string value to the attribute as if `bool:` was not used (attribute always exists in that case).
- The previous bullet point's rules (a la Lit) are fundamentally easier to wrap our head around than what React 19 will have. In React 19, behavior will arbitrarily change depending on whether or not a custom element has a JS property. An example of one of many problems:
  - User is relying on the CSS style `[foo] { color: cornflowerblue }`
    - If the element does not have the `foo` JS property, the style will work.
    - If the element has the `foo` JS property, the style won't work

The approach I've outlined, and which Lit takes, is, simply put, the simplest way to do it, with the least possible confusion, with no compromises and full control of Custom Elements.

## Breaking change

This might be a big change though, so maybe it would need to be shipped with Solid 2.0?

Or, to release something like this in 1.x, there could be a `babel-preset-solid` option to enable the new mode (documenting that if the option is enabled then it will be compiling any 3rd party JSX code with new semantics and thus 3rd party code may break) as well as a new `solid-js/html2` export that has the new `html` behavior (rename it to `solid-js/html` in 2.0).

trusktr:
The more I think about this, the more I dislike that React 19 is going in that direction, it is better than before, but it is still not very good considering that the Lit-based alternative described above is, to be honest, easy and simple.

trusktr:
Oh, hah. And that approach also has a small side effect of increasing performance by eliminating conditional checks like `isCE`. It is negligible, but still.

trusktr:
I made an [Lit playground example](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcbmltcG9ydCB7Y3VzdG9tRWxlbWVudCwgcHJvcGVydHl9IGZyb20gJ2xpdC9kZWNvcmF0b3JzLmpzJztcblxuQGN1c3RvbUVsZW1lbnQoJ3NpbXBsZS1ncmVldGluZycpXG5leHBvcnQgY2xhc3MgU2ltcGxlR3JlZXRpbmcgZXh0ZW5kcyBMaXRFbGVtZW50IHtcbiAgc3RhdGljIHN0eWxlcyA9IGNzc2BwIHsgY29sb3I6IGJsdWUgfWA7XG5cbiAgQHByb3BlcnR5KClcbiAgbmFtZSA9ICdTb21lYm9keSc7XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYFxuICAgICAgPGRpdlxuICAgICAgICBmb289XCJoZWxsb1wiXG4gICAgICAgIC5mb289JHtcImhpXCJ9XG4gICAgICAgIGZvbzI9XCJoZXlcIlxuICAgICAgICAuZm9vMj1cImhvd2R5XCJcbiAgICAgICAgP2Jhcj1cIlwiXG4gICAgICAgID9iYXo9XCJ0cnVlXCJcbiAgICAgICAgP2xvcmVtPVwiZmFsc2VcIlxuICAgICAgICA_ZG9oPSR7dHJ1ZX1cbiAgICAgICAgP2RhaD0ke2ZhbHNlfVxuICAgICAgICBcbiAgICAgICAgZm9vLUJhcj1cInlvXCJcbiAgICAgICAgLmZvby1CYXI9XCJzdXBcIlxuICAgICAgICAuZm9vLUJhcj0ke1wid29yZFwifVxuICAgICAgPlxuICAgICAgICBIZWxsbywgJHt0aGlzLm5hbWV9IVxuICAgICAgPC9kaXY-XG4gICAgYDtcbiAgfVxufVxuXG4vLyBhd2FpdCwgYmVjYXVzZSBMaXQgc3BlY2lmaWNhbGx5IGRvZXMgbm90IHJlbmRlciBhbiBlbGVtZW50J3MgRE9NIHVudGlsIHRoZSBuZXh0IG1pY3JvdGFzay5cbmF3YWl0IG51bGwgICAgXG5cbmNvbnN0IGVsID0gZG9jdW1lbnQuYm9keS5xdWVyeVNlbGVjdG9yKCdzaW1wbGUtZ3JlZXRpbmcnKS5zaGFkb3dSb290LmNoaWxkcmVuWzBdXG4gICAgXG5jb25zb2xlLmxvZyhlbC5nZXRBdHRyaWJ1dGUoJ2ZvbycpKSAvLyBcImhlbGxvXCJcbmNvbnNvbGUubG9nKGVsLmZvbykgLy8gXCJoaVwiXG5jb25zb2xlLmxvZyhlbC5nZXRBdHRyaWJ1dGUoJ2ZvbzInKSkgLy8gXCJoZXlcIlxuY29uc29sZS5sb2coZWwuZm9vMikgLy8gdW5kZWZpbmVkICh0aGF0J3Mgb2RkLCBza2lwIGhhbmRsaW5nIHN0YXRpYyBleHByZXNzaW9uIPCfpKgpXG5jb25zb2xlLmxvZyhlbC5nZXRBdHRyaWJ1dGUoJ2JhcicpKSAvLyBudWxsIChza2lwcyBoYW5kbGluZyBzdGF0aWMgZXhwcmVzc2lvbiDwn6SoKVxuY29uc29sZS5sb2coZWwuZ2V0QXR0cmlidXRlKCdiYXonKSkgLy8gbnVsbCAoc2tpcHMgaGFuZGxpbmcgc3RhdGljIGV4cHJlc3Npb24g8J-kqClcbmNvbnNvbGUubG9nKGVsLmdldEF0dHJpYnV0ZSgnbG9yZW0nKSkgLy8gbnVsbCAoc2tpcHMgaGFuZGxpbmcgc3RhdGljIGV4cHJlc3Npb24g8J-kqClcbmNvbnNvbGUubG9nKGVsLmdldEF0dHJpYnV0ZSgnZG9oJykpIC8vIFwiXCIgKHRydXRoeSwgc28gaXQgZXhpc3RzKVxuY29uc29sZS5sb2coZWwuZ2V0QXR0cmlidXRlKCdkYWgnKSkgLy8gbnVsbCAoZmFsc3ksIHNvIGl0IGRvZXNuJ3QgZXhpc3QpXG4gICAgXG4vLyBPaCwgd2FpdCEgVGhlIG9kZCBvbmVzIGFyZSBiZWNhdXNlIGl0IGNvbmZvcm1zIGNsb3NlbHkgdG8gSFRNTDpcbmNvbnNvbGUubG9nKGVsLmdldEF0dHJpYnV0ZSgnLmZvbzInKSkgLy8gXCJob3dkeVwiXG5jb25zb2xlLmxvZyhlbC5nZXRBdHRyaWJ1dGUoJz9iYXInKSkgLy8gXCJcIlxuY29uc29sZS5sb2coZWwuZ2V0QXR0cmlidXRlKCc_YmF6JykpIC8vIFwidHJ1ZVwiXG5jb25zb2xlLmxvZyhlbC5nZXRBdHRyaWJ1dGUoJz9sb3JlbScpKSAvLyBcImZhbHNlXCJcbiAgICBcbi8vIE5vdGUgdGhhdCBhbGwgdGhyZWUgaGF2ZSBiZWVuIHNldFxuY29uc29sZS5sb2coZWwuZ2V0QXR0cmlidXRlKCdmb28tYmFyJykpIC8vIFwieW9cIlxuY29uc29sZS5sb2coZWwuZ2V0QXR0cmlidXRlKCcuZm9vLWJhcicpKSAvLyBcInN1cFwiXG5jb25zb2xlLmxvZyhlbFsnZm9vLUJhciddKSAvLyBcIndvcmRcIiJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9zaW1wbGUtZ3JlZXRpbmcuanNcIj48L3NjcmlwdD5cbjwvaGVhZD5cbjxib2R5PlxuICA8c2ltcGxlLWdyZWV0aW5nIG5hbWU9XCJXb3JsZFwiPjwvc2ltcGxlLWdyZWV0aW5nPlxuPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMi4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjQuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4zLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ) showing the various cases.

In particular, note the behavior of static attributes (f.e. `.foo="123"`). As noted there, that's because those are verbatim actually-valid HTML attributes, and Lit's `html` template sticks close to HTML standard.

However! I would imagine that sometimes people get confused when `.foo="bar"` does not set the JS property to `"bar"`.

So, we could potentially deviate from this a little bit (but this might mean breaking differences with the type checking and intellisense support from Lit's community), so that `?foo` and `.foo` always have boolean and JS prop behavior regardless if they are static or interpolated. Then there'd always be `attr:` if someone actually intended to set the attribute which I believe would be the rare case. Maybe being *that* close to native HTML is not as ideal.

trusktr:
A downside (but not really, because there'd be no hidden magic), is that for custom elements that map `some-attribute` to a `.someAttribute` property, the person who is using such elements would use either `some-attribute` to set the attribute or `.someAttribute` to set the JS prop. But what you see is exactly what you get! Easy!

trusktr:
I overlooked this was a PR hah. About the change, I think it might be fine. At least for me, in my projects I've been writing all attributes lowercase, and dash cased if multi-word, so I would not run into any issues at all with this change. 

For anyone who may happen to already be relying on the toLowerCase (who would???) it would be breaking. Seems unlikely though? F.e. they write `someThing=` in JSX but it actually sets a `something` JS property.

trusktr:
After some chat with Tito, for hydration there'd need to be a convention for case-sensitive JS props. f.e.:

```html
<some-el this-is-an-AttriBute="no problem" __solid__='{"someProp": 123}'></some-el>
```

trusktr (approved):
I approve because I think there's a low if not zero chance of breaking people with this change (see other comments), and if anything it enables `fooBar=` for setting camelCase JS properties.

This also sparked me to create the following discussion on what I think are the best semantics for attribute/property mappins in the templating system:

- https://github.com/solidjs/solid/discussions/2322

titoBouzout:
I will close this one. Would prefer an update that doesnt do any kind of transformation

PR #365, opened 2024-10-10
fix imported binding ref

mdynnl:
fixes https://github.com/solidjs/solid/issues/2313


ryansolid:
I think this looks good. Atleast from the tests.

PR #366, opened 2024-10-14
fix throwing an error when using `<NoHydration>` before `sharedConfig.context` is initialised

samualtnorman:
We are using Solid to generate internal emails and to avoid the unnecessary `data-hk` attributes I wrap them in `<NoHydration>`. This sometimes caused errors to be thrown so we are using this patch in production to avoid that.


ryansolid:
Thanks

PR #368, opened 2024-10-18
Fix in DOMAttributes

TPReal:
Added the (possibly forgotten?) BoolAttributes and fixed the on:* attributes type..


ryansolid:
Thanks

PR #371, opened 2024-10-21
ssr attribute inlining updates

titoBouzout:
Related to https://github.com/solidjs/solid/issues/2339

ssr: namespaced attributes are inlined into the html in some situations

With config option `ssr:true` sometimes(depending on if `<Dynamic ../>` JSX tag was used or not) the namespaced attributes are being written to the html.

For example `<div attr:name="value"/>`  produces the same thing, instead of `<div name="value"/>` 

Investigating, it seems that `attr:` wasn't the only special namespace with problems. This PR includes updates for avoiding inlining of `prop/attr/bool:` namespaces.

`prop/attr:` works in both cases for my tests (dynamic or not)

There's a need to look into `ssrAttribute` because for some reason `bool:` ends inlined when used as `<div bool:name="quack"/>` (this doesn't happen when used as `<Dynamic component="div" bool:name.../>`). I have looked but didn't feel confident enough to include the namespace in the `ssrAttribute` call, to fix it there.

I'm fine with merging this as is for now

### DIFF
--- a/packages/dom-expressions/src/server.js
+++ b/packages/dom-expressions/src/server.js
@@ -380,8 +380,18 @@ export function ssrElement(tag, props, children, needsId) {
     } else if (BooleanAttributes.has(prop)) {
       if (value) result += prop;
       else continue;
-    } else if (value == undefined || prop === "ref" || prop.slice(0, 2) === "on") {
+    } else if (
+      value == undefined ||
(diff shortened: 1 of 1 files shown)


ryansolid:
Sounds good.. thanks

PR #372, opened 2024-10-21
Add disablepictureinpicture attribute to <video> JSX types

danieltroger:
See https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/disablePictureInPicture

And the list of attributes here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

Screenshot 2024-10-21 at 12 11 58


ryansolid:
Thank you

PR #373, opened 2024-10-21
attributes/properties transformation & custom elements defaulting to attributes

titoBouzout:
It includes two changes:

1. avoids transforming case and dashes for attributes/properties.
2. It doesn't default to `properties` in custom elements. It defaults to `attributes`.

Rationality: Seems like numerous people expect custom elements to just receive attributes  [1](https://discord.com/channels/722131463138705510/1297146159202046003), [2](https://github.com/solidjs/solid/issues/2339), [3](https://github.com/solidjs/solid/discussions/2322#discussion-7274338), [4](https://github.com/solidjs/solid/issues/2321), and they will use `prop:` when they want to set a property. It seems strange wanting to use `prop:`, but it looks to me, they are used to dealing with this sort of problem. Kind of makes sense as attributes can only receive strings.

Avoiding transformation of case and dashes, makes the issue easier to deal with it. I do not think this will cause any problem, as one won't mix up case when using attr/props, so normalizing it is unnecessary and could cause some other [issues](https://github.com/ryansolid/dom-expressions/pull/351) 

If I do not miss anything, this only affects custom-elements.  I suggest merging it whenever possible, as this is something people seem to expect to behave this way. I also would like to recommend merging it as `v1.10.0` to not break and alienate people using the previous behaviour.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
@@ -224,14 +224,6 @@ export function toEventName(name) {
   return name.slice(2).toLowerCase();
 }
 
-export function toAttributeName(name) {
-  return name.replace(/([A-Z])/g, g => `-${g[0].toLowerCase()}`);
-}
-
-export function toPropertyName(name) {
-  return name.toLowerCase().replace(/-([a-z])/g, (_, w) => w.toUpperCase());
-}
-
 export function wrappedByText(list, startIndex) {
   let index = startIndex,
     wrapped;
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -219,7 +219,7 @@ export function assign(node, props, isSVG, skipChildren, prevProps = {}, skipRef
   for (const prop in prevProps) {
     if (!(prop in props)) {
       if (prop === "children") continue;
-      prevProps[prop] = assignProp(node, prop, null, prevProps[prop], isSVG, skipRef, props);
+      prevProps[prop] = assignProp(node, prop, null, prevProps[prop], isSVG, skipRef);
     }
   }
   for (const prop in props) {
@@ -228,7 +228,7 @@ export function assign(node, props, isSVG, skipChildren, prevProps = {}, skipRef
       continue;
     }
     const value = props[prop];
-    prevProps[prop] = assignProp(node, prop, value, prevProps[prop], isSVG, skipRef, props);
+    prevProps[prop] = assignProp(node, prop, value, prevProps[prop], isSVG, skipRef);
   }
 }
 
@@ -260,7 +260,9 @@ export function getNextElement(template) {
   if (!hydrating || !(node = sharedConfig.registry.get((key = getHydrationKey())))) {
     if ("_DX_DEV_" && hydrating) {
       sharedConfig.done = true;
-      throw new Error(`Hydration Mismatch. Unable to find DOM nodes for hydration key: ${key}\n${template ? template().outerHTML : ""}`);
+      throw new Error(
+        `Hydration Mismatch. Unable to find DOM nodes for hydration key: ${key}\n${template ? template().outerHTML : ""}`
+      );
     }
     return template();
   }
@@ -321,18 +323,14 @@ function isHydrating(node) {
   return !!sharedConfig.context && !sharedConfig.done && (!node || node.isConnected);
 }
 
-function toPropertyName(name) {
-  return name.toLowerCase().replace(/-([a-z])/g, (_, w) => w.toUpperCase());
-}
-
 function toggleClassKey(node, key, value) {
   const classNames = key.trim().split(/\s+/);
   for (let i = 0, nameLen = classNames.length; i < nameLen; i++)
     node.classList.toggle(classNames[i], value);
 }
 
-function assignProp(node, prop, value, prev, isSVG, skipRef, props) {
-  let isCE, isProp, isChildProp, propAlias, forceProp;
+function assignProp(node, prop, value, prev, isSVG, skipRef) {
+  let propAlias, forceProp;
   if (prop === "style") return style(node, value, prev);
   if (prop === "classList") return classList(node, value, prev);
   if (value === prev) return prev;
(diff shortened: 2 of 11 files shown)


packages/dom-expressions/src/client.js
```diff
@@ -366,15 +364,16 @@ function assignProp(node, prop, value, prev, isSVG, skipRef, props) {
...
   ) {
     if (forceProp) {
       prop = prop.slice(5);
       isProp = true;
     } else if (isHydrating(node)) return value;
     if (prop === "class" || prop === "className") className(node, value);
-    else if (isCE && !isProp && !isChildProp) node[toPropertyName(prop)] = value;
-    else node[propAlias || prop] = value;
+    else if (isCE && !isProp && !isChildProp) {
+      node.setAttribute(prop, value);
```
trusktr, line 375: This is nice, unlike React's which forces a string coercion of `value` before passing it to `setAttribute`. The native DOM `setAttribute` method already will do that coercion, but! a custom element class can override the `setAttribute` method to also handle non-string values and for example map them directly to JS properties, which is not possible with React at time of this writing.

This means that, even if Solid JSX calls `el.setAttribute`, a custom element can itself be designed to pass the value directly to JS and skip the native `setAttribute`, therefore achieving the same thing as `prop:` without the user having to use `prop:` syntax. The A-Frame library, for example, overrides `setAttribute` like this for all its elements. I'm contemplating adding this to Lume Element. In my case, I could map JS value types to the JSX attribute types. But other people's CE libraries don't have to, they can map only string types for non-`prop:` attributes for example.

trusktr: In fact, had React at least removed that stupid string coercion, they'd have largely eliminated the need for React 19 CE support because all CEd would have had a way to accept non-string values.

But they didn't listen.

packages/dom-expressions/src/client.js
```diff
@@ -366,15 +364,16 @@ function assignProp(node, prop, value, prev, isSVG, skipRef, props) {
...
+    (isCE = node.nodeName.includes("-") || "is" in props)
   ) {
     if (forceProp) {
       prop = prop.slice(5);
       isProp = true;
     } else if (isHydrating(node)) return value;
     if (prop === "class" || prop === "className") className(node, value);
-    else if (isCE && !isProp && !isChildProp) node[toPropertyName(prop)] = value;
-    else node[propAlias || prop] = value;
+    else if (isCE && !isProp && !isChildProp) {
```
trusktr, line 374: Seems like we wouldn't need `isCE` anymore. The logic would be for any element (doesn't matter if it is CE or not) set the attribute (except for the special cases).

packages/dom-expressions/src/client.js
ryansolid: Overriding `setAttribute` is interesting and something I hadn't thought of. I'm sure it has other implications/inconsistency but it definitely is interesting.

packages/dom-expressions/src/client.js
ryansolid: Yeah that is the benefit here. Not treating CE's special. I think that is directionally where we go. I sort of over compensated because I wanted WC development to be pleasant but we shouldn't really be trying to pave over the awkward parts. They are the way they are and its up to the specs to handle that.

titoBouzout:
Just as a reminder, it may be a need to also update `component-register` for `solid-element` 
https://github.com/ryansolid/component-register/blob/master/src/utils.ts#L141-L152

ryansolid:
So are @titoBouzout and @trusktr in agreement? If you guys are good with this we can merge it into the next branch. This won't be Solid 1.0 time period but lets start 2.0 right.

titoBouzout:
@ryansolid I spoke with @trusktr a few days ago asking for this, and we are in agreement.

ryansolid:
Awesome thank you. Merged into next.

packages/dom-expressions/src/client.js
trusktr: > Overriding `setAttribute` is interesting

Indeed, for example:

```js
class MyEl extends HTMLElement {
  setAttribute(attr, value) {
    if (typeof value === string) super.setAttribute(value) // attributeChangedCallback can handle parsing, or etc.
    else this[attr] = value
  }
}
```

but React had get in the way by doing this:

```js
theElement.setAttribute(jsxProp, String(theValueFromJSX))
```

forcing strings on everyone for no reason at all.

Luckily React 19 is *finally* out now and things are better.

trusktr:
Is this coming out in Solid 2.0? Or learlier? Concerned about how breaking it is.

What are the steps to try this with Solid 1.x? I can report back if anything breaks.

titoBouzout:
> Is this coming out in Solid 2.0? Or earlier? I'm wondering how breaking it is.

This is coming out in Solid 2.0, not earlier, as we decided to introduce the breaking changes in one step instead of gradually (to make it less confusing). There are some deprecations notices tho.

For a custom-element is pretty breaking because we have been coding custom elements with properties instead of attributes. So unless these been reflecting everything and also not using the prop transformation `cash-case` to `camelCase` it will be most likely be broken.

> What are the steps to try this with Solid 1.x? I'm guessing I'll just `npm link` it into an `npm link`ed solid-js, and build solid-js with it.

There is `v2.0.0-experimental.0` https://github.com/solidjs/solid/releases/tag/v2.0.0-experimental.0 
but be aware that some set of patches for everything related to attribute/properties are missing. (I will see of sending these this week) 

Basically what we agreed and should expect at release is: 
- set everything as attributes except for some special properties (innerHTML, textContent, and friends) [so attributes are lowercase, no more `autoplay` vs `autoPlay`]
- use `prop:` for properties 
- event handlers in `cameCase` (with native handlers as `on:`)
- no props/attrs transformation

Think this is a good middle ground and we are pretty happy about that, right?

packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
```diff
@@ -224,14 +224,6 @@ export function toEventName(name) {
   return name.slice(2).toLowerCase();
```
trusktr, line 224: Having just `name.slice(2)` here would be nice. Then event delegation will work even for custom events regardless of the name right?

ALong with that, what about a way to configure events, f.e. `onMyEvent={[e => console.log(e), {bubbles: false, delegated: true}]}` or `onotherEvent={{handleEvent: e => console.log(e), bubbles: false, delegated: true}}`

trusktr:
> * set everything as attributes except for some special properties (innerHTML, textContent, and friends)

I'd personally be ok using only `prop:` for these, f.e. `prop:innerHTML`, `prop:textContent`, making it crystal clear.

PR #374, opened 2024-10-23
loading lazy with dynamic value

titoBouzout:
Related to https://github.com/solidjs/solid/issues/1828#issuecomment-2429672721

Always use `importNode` bit for `img` and `iframe` with attribute `loading` regardless of its value.


ryansolid:
Thanks

PR #375, opened 2024-10-23
add support for MathML

titoBouzout:
Adds support for MathML tags without trying to change things much or get too fancy. 

It tests the template content for MathML tags just before doing the call expression and adds a new argument `isMathML` to the client `template` function.

I haven't actually tested the client side of it, other than making sure it works in the console. I am somewhat confident about it. The `template` tag doesn't have a `content` when created with a namespace, kind of makes sense.. but not really!   
 
While MathML sort of works already, it has the problem that when children are dynamic these are created with the wrong namespace and break. This should fix it.


titoBouzout:
This is motivated by https://discord.com/channels/722131463138705510/817960620736380928/1298475473898311701

titoBouzout:
closed by missclick, sorry, I am attemping to look into the conflicts!  Thanks

titoBouzout:
I am highly interested for this to be merged (whenever possible of course)

Do you forsee any issue with this?

ryansolid:
It's funny. I'm trying to remove all the arguments to `template` function not add more. How many of these are there? Part of me wonders if custom renderers are the right approach after a while. SVG was pain, custom-elements more,.. is mathML the last of this? 

I see template handling, but spreads also need special handling I think. I'd look anywhere there is `isSVG` in the client.js of DOM Expressions. We probably need to add specific constants for it too so it can be looked up at runtime.

titoBouzout:
In theory, we should be using `importNode` but it has the performance hit that `cloneNode` doesn't have. So that's an argument that could possibly go away after chrome fixes the performance issue. 

About `isSVG` and `isMathML` I understand we do not aim to be a fully compatible XML renderer, but HTML has these two namespaces that some people use and in my opinion need to be supported. 

The way I implemented this in pota is by having a `xmlns` argument instead, setting a context when the `xmlns` changes, query the context and using `createElementNS` in the template if necessary. So the `xmlns` argument is only used by the top-most element that changes the namespace and children get their namespace from the context.  Dev tools report a 1% performance hit of self-time.

I'd prefer the context approach rather than the PR I sent because: it's simpler, it's generic (supports different namespaces in 1 implementation), its future-proof (if more tags are added these will be automatically supported), and It's somewhat warranted to work in the different dynamic cases.

ryansolid:
I don't know if the context solution is simpler. I think it is better but it also requires new capabilities in DOM expressions to inject context and do lookups. It does address my concern with this being endless though but it is a much bigger lift. I don't see that getting in 1.0. Honestly it isn't really high enough priority for 2.0. I think we bite the bullet and just merge this. It will be the last one for a while it sounds like.

PR #382, opened 2024-11-11
handle moving nodes

### LINKED ISSUE #2357: Cannot reuse a dom element when switching mount locations
I'm trying to reuse the same dom element, and attach it in different places in the ui.
The element needs to be kept alive and animated from place to another.
But even though that element is rendered only in one place at a time,
solid's rendering of it is not stable.
Sometimes the element is added correctly, sometimes it's removed from the dom completely and sometimes it appears in the wrong spot.

```tsx
const my_dom_element = document.createElement('div')

return <>
	<div>
		<Show when={something() === 'foo'}>
			{my_dom_element}
		</Show>
	</div>
	<div>
		<Show when={something() === 'bar'}>
			{my_dom_element}
		</Show>
	</div>	
</>
```

Playground link:
https://playground.solidjs.com/anonymous/d2ce885b-d473-4641-94a6-d9c659bc1e22

mdynnl:
fixes https://github.com/solidjs/solid/issues/2357

### DIFF
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -460,7 +460,6 @@ function insertExpression(parent, value, current, marker, unwrapArray) {
   if (value === current) return current;
   const t = typeof value,
     multi = marker !== undefined;
-  parent = (multi && current[0] && current[0].parentNode) || parent;
 
   if (t === "string" || t === "number") {
     if (hydrating) return current;
@@ -570,7 +569,14 @@ function appendNodes(parent, array, marker = null) {
   for (let i = 0, len = array.length; i < len; i++) parent.insertBefore(array[i], marker);
 }
 
+let moved;
+
 function cleanChildren(parent, current, marker, replacement) {
+  if (!moved) {
+    moved = new WeakSet();
+    queueMicrotask(() => (moved = null));
+  }
+
   if (marker === undefined) return (parent.textContent = "");
   const node = replacement || document.createTextNode("");
   if (current.length) {
@@ -579,9 +585,12 @@ function cleanChildren(parent, current, marker, replacement) {
       const el = current[i];
       if (node !== el) {
         const isParent = el.parentNode === parent;
-        if (!inserted && !i)
-          isParent ? parent.replaceChild(node, el) : parent.insertBefore(node, marker);
-        else isParent && el.remove();
+        if (!inserted && !i) {
+          node.parentNode && moved.add(node);
+          isParent && !moved.has(el)
+            ? parent.replaceChild(node, el)
+            : parent.insertBefore(node, marker);
+        } else isParent && el.remove();
       } else inserted = true;
     }
   } else parent.insertBefore(node, marker);
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -569,14 +569,17 @@ function appendNodes(parent, array, marker = null) {
   for (let i = 0, len = array.length; i < len; i++) parent.insertBefore(array[i], marker);
 }
 
-let moved;
+const moved = new Set();
+let scheduled = false;
 
 function cleanChildren(parent, current, marker, replacement) {
-  if (!moved) {
-    moved = new WeakSet();
-    queueMicrotask(() => (moved = null));
+  if (!scheduled && moved.size) {
+    queueMicrotask(() => {
+      moved.clear();
+      scheduled = false;
+    })
+    scheduled = true;
   }
-
   if (marker === undefined) return (parent.textContent = "");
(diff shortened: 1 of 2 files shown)


packages/dom-expressions/src/client.js
```diff
@@ -460,7 +460,6 @@ function insertExpression(parent, value, current, marker, unwrapArray) {
   if (value === current) return current;
   const t = typeof value,
     multi = marker !== undefined;
-  parent = (multi && current[0] && current[0].parentNode) || parent;
```
mdynnl, line 463: Not sure what might break by this removal but this for sure breaks the parent comparison in `cleanChildren`

packages/dom-expressions/src/client.js
```diff
@@ -570,7 +569,14 @@ function appendNodes(parent, array, marker = null) {
   for (let i = 0, len = array.length; i < len; i++) parent.insertBefore(array[i], marker);
 }
 
+let moved;
+
 function cleanChildren(parent, current, marker, replacement) {
+  if (!moved) {
+    moved = new WeakSet();
+    queueMicrotask(() => (moved = null));
+  }
```
mdynnl, line 578: `WeakSet` is already used by hydration so doesn't hurt anything but there might be better approaches than this.

packages/dom-expressions/src/client.js
titoBouzout: Ryan made a comment about this code in a related issue here https://github.com/ryansolid/dom-expressions/pull/307#issuecomment-1904967433 
I will have to study it again to refresh my mind, I'm not sure I actually ever understood that part

packages/dom-expressions/src/client.js
titoBouzout: I suggest considering a strong referenced array to be cleared in the microtask as `moved.length = 0`. As the cost of instantiating could be removed, and also a set will try to deduplicate the nodes

mdynnl: I was thinking if weakset was faster than array to check for element existence. Any form of book-keeping works really. We could also keep it in the nodes but I don't know the performance side of this. Essentially we need to know if a node has changed position in the tree.

packages/dom-expressions/src/client.js
mdynnl: I've tried to track this down and, oh well 😅, no wonder even Ryan can't remember this. This check existed since the beginning.

ryansolid: Yeah someone made a codesandbox that was sorting nodes around randomly on an interval and it would occasionally break. I added that way back. But I think its need predates the "modern" way we handle fragments I changed to back in Feb 2019. 

In any case I'm moving forward with this not there right now in 2.0 stuff.

ryansolid:
In general i'm interested in this but it is likely not the sort of thing I can see being changed in Solid 1.0 lifespan. Already making moves towards 2.0's structure which gives us ample time to performance benchmark and vet if this makes sense.

What this means for the Dom Expressions project is more than likely a bit of a jump in version 0.40.0 we can look at including it there. It might be possible that it is time for DOM Expressions to go to 1.0 given how big of a change is going on.

ryansolid:
Hi @mdynnl — thanks for putting this together. Flagging a few things from tracing the canonical case in solidjs/solid#2357 before this lands.

The bug-triggering path for <Show when={x}>{sharedNode}</Show> (and the variants in #2030) routes through reconcileArrays, not cleanChildren. In multi-marker mode insert seeds current=[placeholderText], so the first truthy update is reconcileArrays(parent, [placeholder], [sharedNode]), and the destructive ops at risk are the fallback-branch a[aStart++].remove() / parent.replaceChild(b[bStart++], a[aStart++]) plus the after = a[aEnd-1].nextSibling read when a[aEnd-1] has migrated. This PR doesn't see that path.

The moved Set is also microtask-scoped, but a migration is structural — it persists with the node until something explicitly re-inserts it. Anything that crosses a microtask boundary (or where the dependent slot updates in a separate flush) sees an empty Set and treats the migrated node as still-owned. A per-node ownership tag (Symbol-keyed reference to the owning slot's marker, written on every runtime insertion site) handles both timing and cross-microtask cases.

The included tests both use r.insert(…) without threading current back in, so each call sees current=[] and dispatches to appendNodes — they validate appendNodes+cleanChildren cooperation but not the reconcile-driven migration.

I'm exploring the tag-based approach on next.

ryansolid:
Thank you for the PR. THis was very helpful pushing this issue forward and I am very appreciative of yours and @titoBouzout persistence here. This feature has landed now and is released in Solid in beta.14.

Thank you.

Issue #384, opened 2024-11-26
`playsinline` attribute is set wrong in hydrated output version


danieltroger:
What happens: `videoElement.playsinline === true` but the browser only respects `videoElement.playsInline === true`

To fix, solid-js should set `videoElement.playsInline` instead.

To reproduce: go to https://playground.solidjs.com/anonymous/65323853-6f9d-4954-9b03-2ffd622769ce and select "output"

You will see the output is correct for "client side rendering"

<img width="789" alt="Screenshot 2024-11-26 at 11 46 45" src="https://github.com/user-attachments/assets/a9e4c40c-dd54-47c3-a829-5267cc6cebf7">

Now, at the bottom, switch compile mode to "Client side rendering with hydration"
<img width="655" alt="Screenshot 2024-11-26 at 11 48 17" src="https://github.com/user-attachments/assets/a9bebc50-b1c4-4d22-99d3-05ecdb19e084">

You will now notice that solid-js changes the wrong property on the video element
<img width="783" alt="Screenshot 2024-11-26 at 11 47 37" src="https://github.com/user-attachments/assets/c14314e2-75da-4e57-8579-a111c64e2e20">

titoBouzout:
I sympathise with the expectation, but the cost of such expectation is high. I will list a few arguments:

1. Transforming lowercase attributes to their property versions means we will need to ship the map of attributes -> properties 
2. Transforming the case of attributes/properties leads to problems, which is something I have been working on trying to avoid https://github.com/ryansolid/dom-expressions/pull/373
3. This case of setting the attribute directly to boolean is something I have been trying to push against as seen in https://github.com/ryansolid/dom-expressions/pull/380 . We can make an update to dom-expressions that native booleans (a boolean not coming from a signal) to be inlined at compiler time. So for the case at hand, it would have been transformed to `<video playsinline=true>` directly client and server side. Without the need to `setAttribute/setProperty`. It is preferred if you use the string version when the value is hardcoded like that, as its inlined.
4. Pretending an attribute is a property is misleading. It would be much better if you use thee camelCase version when you actually want to set the property (as I think it should be). Think we can use properties instead of attributes when that's the case for custom elements without having to use `prop:`

That's what just came to mind

danieltroger:
Thank you for the quick answer!

> I sympathise with the expectation, but the cost of such expectation is high.

[...]

PR #385, opened 2024-11-28
inline in template boolean literals

titoBouzout:
This change, inlines in the template boolean literals. For example

`<el attr={true}/>` becomes  `<el attr/>`
and 
`<el attr={false}/>` becomes  `<el/>`

Considerations:
- this change skips doing this transformation for "custom elements", because as we are defaulting to properties I suspect something will break.
- I do not think this is a breaking change, at least from my tests I didn't see anything that could cause problems.
- this is related to an effort to fix the attributes vs properties problem

The most recent issue illustrating this props vs attr problem comes from https://github.com/ryansolid/dom-expressions/issues/384 In there they are doing `<video playsinline={true} />;`, in production code currently:
- client side works because compiles it to `_el$.playsInline = true;`
- in ssr is inlined in the template
- ssr with hydration fails because it tries to add the property with the lowercase name as `_$setProperty(_el$, "playsinline", true);`

With this change, 
- boolean attributes will always be inlined for the `true` case and skipped for the `false` case
- once the pr with the types gets merged it will only accept `"true" | boolean`, so trying to do `"false"`` as a string should be a type error

I am hoping before solid 2.0 we can get rid of the hardcoded list of properties 

Thanks!

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -487,7 +487,12 @@ function transformAttributes(path, results) {
       if (
         t.isJSXExpressionContainer(value) &&
         (reservedNameSpace ||
-          !(t.isStringLiteral(value.expression) || t.isNumericLiteral(value.expression)))
+          !(
+            t.isStringLiteral(value.expression) ||
+            t.isNumericLiteral(value.expression) ||
+            // remove `!isCE` when custom elements start defaulting to attributes
+            (t.isBooleanLiteral(value.expression) && !isCE)
+          ))
       ) {
         if (key === "ref") {
           // Normalize expressions for non-null and type-as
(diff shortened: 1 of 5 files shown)


ryansolid:
The thing is the playsinline probably should be mapped properly anything that is denoted as a property should have translation for attributes. We just missed this one.

titoBouzout:
> Only question here is does this PR consider the fake booleans? Like things that need to be string "false" vs "true" where the application of just the property being there doesn't work. We have the constants list to differentiate between those.

Yes, this PR only affects literal `true` and `false`, it doesn't change enumerated attributes such `"yes"`, `"no"`, `"true"`, `"false"`, there are tests for such a case.

> Like to be clear. It is very intentional to use the properties for booleans and if we are to be able to set them as attributes as well then there is no way to get rid of the lookup.

Can you please clarify what do you mean with its intentional?

ryansolid:
Sorry for the delay of response. What I mean is that true boolean attributes.. like readonly or hidden etc.. are intended to be set as properties when dynamic in the client. Unlike say `draggable` which is set as an attribute.

I see this is a static case so we can inline them regardless.

But I was saying this sounds like a bug:

> ssr with hydration fails because it tries to add the property with the lowercase name as `_$setProperty(_el$, "playsinline", true);`

It should be properly cased. Sure it is static here so it could be inlined but it should be set as `_$setProperty(_el$, "playsInline", true);` if it were dynamic.

I guess I can merge this as is.

PR #386, opened 2024-12-03
update types, deprecate lowercase event handlers

titoBouzout:
It has been noted that supporting both `onclick` and `onClick` makes  components a bit more complicated to type and support, as a consumer could use any of both. Literally copying and pasting the issue

> oh it's just the same old issue: If JSX.ButtonHTMLAttributes<HTMLButtonElement> has both onclick and onClick, then when writing a component to wrap it, you either have to pick one or the other Omit<JSX.ButtonHTMLAttributes<HTMLButtonElement>, "onClick"> or you have to support both. Consider

```js
function PartyButton(props: JSX.ButtonHTMLAttributes<HTMLButtonElement>) {
  let button: HTMLButtonElement | undefined;
  return (
    <button
      {...props}
      ref={fwd(props, ref => (button = ref))}
      onclick={e => (
        triggerConfettiAnimation(button),
        typeof props.onclick === "function" ? props.onclick?.(e) : undefined
      )}
    />
  );
}
```

> now when someone uses this `<PartyButton onClick={() => console.log("clicked")}>Example</PartyButton>` the `onClick` doesn't get called

From https://discord.com/channels/722131463138705510/817960620736380928/1313201415182745630 by @webstrand

---

Kobalte and corvu both use camelCase events. 

[The eslint plugin](https://github.com/solidjs-community/eslint-plugin-solid/blob/main/packages/eslint-plugin-solid/docs/event-handlers.md) also prefers onClick over onclick, and by default will autocorrect accordingly. (But there's an option to ignoreCase) [from @edemaine ]


ryansolid:
It is interesting Svelte opted into all lowercase event names when they got rid of `on:click`. It is an interesting choice to default lower case for all HTML things and not for events. I think the argument is that they aren't native events. Where as `on:____` for us are. 

At one time we had a difference between casing.. `onClick` would do what we do now and `onclick` would just set the attribute like native elements. I think I like the camelCase because it is clear that we aren't expecting a string. All the DOM element ones expect a string. It means if people copy and paste from HTML it will type error but that is good because we don't do string event handlers. 

Yeah I think this is a good addition to the other one.

titoBouzout:
One very interesting feedback on why event listeners are `camelCase` and not lowercase or case-sensitive is in https://discord.com/channels/722131463138705510/817960620736380928/1340489565508472886
invitation if you need https://discord.gg/solidjs

Issue #337, opened 2024-08-06
Issue with webcomponents that have HTML Comment elements


spion:
Some webcomponents (like Ionic's ion-label) generate additional comment children

For those, the code frragment

https://github.com/ryansolid/dom-expressions/blob/8d2d82fd96a08c82b933fcf87ef521c4b54cf490/packages/dom-expressions/src/client.js#L428

is going to try and update the comment rather than the actual text content.

A repro that isn't very minimal: https://stackblitz.com/edit/github-pyx5ga?file=package.json,app%2Fclient.tsx,vite.config.js,index.html,app%2Fionic-css.ts

You will notice that `<IonLabel />` generates empty comment elements, which in turn cause SolidJS to update them if the only child of the component is a string

ryansolid:
Our system works by walking known nodes, I don't don't know if there is anything to be done here. I'm open to suggestions but if something changes the shape of the HTML from what the JSX outputs we aren't going to know where things are.

titoBouzout:
Yes, we can do something. The fatal flaw is the assumption that the node is "ours". Instead of assuming the node it's ours we can save in `current` the actual node. `insertExpression` shouldn't make any assumption, because doing so will break any code that changes the tree. 

I am not trying to be far-fetched, what I mean is that as long as we have references to nodes, and as long as "others" do not mess up with "our" stuff(nodes), then we can keep doing our thing without breaking anything.

mdynnl:
We can address this [here](https://github.com/ryansolid/dom-expressions/pull/382) along with similar cases. The necessary change also shouldn't have that much impact on perf as we just need to flip it to make the node "ours".

titoBouzout:
That would be nice, I only identify these two lines that could become problematic. In theory, it should be a matter of saving in `current` the text nodes, instead of the text, and then making sure it compares to that, I see in the `multi` case sort of already does.  This function is a bit complex

https://github.com/ryansolid/dom-expressions/blob/a1d24a1a4fb07c3a63919c978ec4180896de9248/packages/dom-expressions/src/client.js#L479-L480

ryansolid:
Yeah performance is a big deal here. Otherwise everything would be multi. By assuming we render all children shortcuts a bunch of operations. That being said I believe this issue is different than the moving nodes in the issue.

titoBouzout:
Just had an idea, the original code:

```js
if (current !== "" && typeof current === "string")
  current = parent.firstChild.data = value;
else 
[...]

PR #388, opened 2024-12-08
non-destructive renderer

mdynnl:
This PR attempts to find a happy path to render without destroying unknown nodes as seen in https://github.com/ryansolid/dom-expressions/issues/337

I also added a playground to easily test the changes in the browser. `pnpm --filter playground dev`

### DIFF
--- a/apps/playground/index.html
+++ b/apps/playground/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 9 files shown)


ryansolid:
Closing with #337. The insert path assuming it owns the child list is a judgement call, not a bug: if Solid moves the node it is our responsibility; if a web component (or anything else) injects comments/nodes, it is not. Walking foreign children on the happy path is too expensive to take.

PR #389, opened 2024-12-09
Treat tagged templates as function calls

edemaine:
Fixes https://github.com/solidjs/solid/issues/2377

Without the change, the test compiles to the incorrect

```js
const singleTemplateLiteral = greeting`world`
```


ryansolid:
Awesome thanks.

PR #398, opened 2025-01-24
makes ssr consistent with templates by avoiding adding `=""` to empty attributes 

### LINKED ISSUE #391: Inconsistent SSR of Boolean attributes
Consider the following example code:

```tsx
function Attr() {
  return (
    <>
    <input checked={true} />
    <input checked />
    </>
  );
}
```

In SSR mode, this gets transpiled to:

```js
import { ssr as _$ssr } from "solid-js/web";
import { ssrHydrationKey as _$ssrHydrationKey } from "solid-js/web";
var _tmpl$ = ["<input", " checked=\"\">"],
  _tmpl$2 = ["<input", " checked>"];
function Attr() {
  return [_$ssr(_tmpl$, _$ssrHydrationKey()), _$ssr(_tmpl$2, _$ssrHydrationKey())];
}
```

While technically correct according to [MDN](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML), the two templates should not differ. I'd expect `<input checked={true} />` to also result in the template  `["<input", " checked>"]` as opposed to the actual `["<input", " checked=\"\">"]`.

Furthermore, if the value of the `checked` property is computed at runtime, i.e., `<input checked={Date.now() > 0} />`, the rendered HTML string also is `<input checked>`, inconsistent with the `<input checked="">` generated by the transpiler when the value is a constant.

titoBouzout:
It avoids in templates(in a few places on where still added `=""`) and SSR having an empty attribute as `<a attr=""/>`, instead it just adds the attribute as `<a attr/>`. 

It fixes #391


ryansolid:
Thanks

PR #400, opened 2025-02-14
better intellisense for target, controlslist and type

huseeiin:
added common values for \<media> controlslist

added common values for \<input> type

added common values for \<a> target

the rest is prettier formatting


ryansolid:
Thanks

PR #409, opened 2025-04-16
main: undo inline in template boolean literals #385

titoBouzout:
Undo #385 that inadvertently introduced a breaking change for 1.9.x, by changing the behaviour of things like `<div draggable={true}/>` to be `<div draggable/>` instead of the current `<div draggable="true"/>`., this also affects other attributes as `aria-hidden`, etc.

However, the change undid here is already applied and desired in `next`, as that's the behaviour we want for Solid 2.0


ryansolid:
Now making sure that next doesn't end up with this undo downstream is going to be a bit fun after my next merge.

titoBouzout:
@ryansolid Yeah, but do not worry, if that happens, let me know(I will be watching anway), Ill re-apply the new behaviour that is very small, and the test file..  that's a matter of a bit of time on them. Thanks!

PR #410, opened 2025-04-20
main: update JSX types

titoBouzout:
Updates `main` types for solid 1.9.x

- Backports non-breaking changes/updates from `next`
- add node script to sync `jsx-h.d.ts` and `jsx.d.ts` with instructions on readme about how to update types
- add MathML types
- add/update several attributes/properties/events/interfaces
- fix event handlers code being outdated and out of sync for `jsx-h.d.ts` with `jsx.d.ts`
- add some new aria attributes
- add `FunctionMaybe` in some missing places
- sort properties for elements to make it easier to merge and compare with other versions of this file
- add missing event handlers in body/svg/iframe
- add `clip-path` to some SVG elements

So from now on, to update JSX types:

1. we edit `jsx-h.d.ts` (the source of truth)
2. run `node jsx-update.mjs` in `packages/dom-expressions/src/` which will write a new `jsx.d.ts` with the irrelevant types stripped.

The script to transform the `d.ts` file is not pretty (is not an AST transformation) but its simple enough and works for me.

The difference between `main` and `minor` types are:
1. `minor` includes deprecation notices for camelCase attributes and lowercase event handlers, such `/** @deprecated Use lowercase attributes */`
2. boolean types in `main` are `boolean | undefined`  while in `minor` are `"" | boolean | undefined` (this may change in minor soon, but just informing what are the differences)


ryansolid:
THank you

PR #412, opened 2025-04-24
Pass property names as quoted strings.

### LINKED ISSUE #411: Properties with `-` in their name (kebab-case) cause a compile error in code output from lit-dom-expressions
This occurs for custom components only; when rendering an HTML element it works as expected.

Minimum reproduction: https://playground.solidjs.com/anonymous/9321a037-f6e4-41cc-aa8e-7250323de592

The error looks something like
```
Uncaught (in promise) SyntaxError: missing : after property id
    functionBuilder html.js:225
    parseTemplate html.js:661
    createTemplate html.js:256
    html html.js:675
```
in Firefox (as bundled in solidjs/html, of course).

afranchuk:
Fixes #411.


ryansolid:
Thanks

PR #413, opened 2025-04-27
fix prop/attr namespace error on children when value is a string

titoBouzout:
The following worked: 

```js
const div = <div prop:poster="test"></div>;
```
https://playground.solidjs.com/anonymous/4ecb5e58-9d49-48cd-b088-cd01116b24d1

but the following gives an error

```js
const div = <div><div prop:poster="test"></div></div>;
```
https://playground.solidjs.com/anonymous/d717fcfe-1aca-48fa-ba89-5f38fa6f8d91

--- 

Was hard to trace it down but figured the problem was `results.id` being undefined because of `info.skipId`  which calls `detectExpressions`.  

`detectExpressions` seems to try to skip when is a string/number.  

https://github.com/ryansolid/dom-expressions/blob/main/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js#L1105-L1110
 
I saw there's a case for the "use" namespace, so added the others too `attr/prop/bool` and that seems to have worked.

edit: removed `bool` from it as its inlined so not needed.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -1106,7 +1106,6 @@ function detectExpressions(children, index, config) {
             ["textContent", "innerHTML", "innerText"].includes(attr.name.name) ||
             (attr.name.namespace &&
               (attr.name.namespace.name === "use" ||
-                attr.name.namespace.name === "bool" ||
                 attr.name.namespace.name === "attr" ||
                 attr.name.namespace.name === "prop")) ||
             (t.isJSXExpressionContainer(attr.value) &&
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -457,6 +457,64 @@ function transformAttributes(path, results) {
 
   let needsSpacing = true;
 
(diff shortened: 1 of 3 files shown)


ryansolid:
Its because they should get inlined I believe. `prop` can't for obvious reasons but `attr` or `bool` should be fine. If they are inlined we don't need to walk to them so we should be able skip ids.

titoBouzout:
Yes, thanks, for future reference `results.id` is a DOM element. What `skipId`  is trying to do, is avoiding the unnecessary refs

```js
const template59 = (() => {
  var _el$23 = _tmpl$22();
  return _el$23;
})();
```

and instead just have it as 

```js
const template59 = _tmpl$22();
```

--- 

OK, so I removed `attr:` from the list of namespaces, but it kept crashing.  It didn't seem to inline `attr:` when the value is `string/numeric`.  So I moved the code that inlines values for regular attributes to the parent scope, and reused it in an `attr:` case for when that's possible (value is string/numberic).

Notes:
- had to scope/hoist that function because it uses the "local" `needsSpacing`
- had to skip inlining `attr:onclick` because this shift an incredible amount of tests, as there's an early `const template12 = <div attr:onclick="console.log('hi')"/>` somewhere.  I have been working on attributes/properties and don't want to mess up the diffs with irrelevant number swapping.

ryansolid:
Ahh I see.. awesome. Thanks.

PR #414, opened 2025-05-02
main: update constants

titoBouzout:
Updates DOM constants in branch `main`

- add some new boolean properties/attributes (including missing `allowFullscreen` , `defer`)
- update prop aliases to add new properties  (including missing `allowFullscreen` )
- add `webview` to html tags
- add some missing deprecated tags (as the list includes deprecated tags, I added the ones I know are missing, following suit)

Note: if the PR looks good to you, can you please let me know before merging, so I sort the lists, I avoided sorting the lists because the diffs become harder to read.

Thanks!

### DIFF
--- a/packages/dom-expressions/src/constants.js
+++ b/packages/dom-expressions/src/constants.js
@@ -1,16 +1,17 @@
+// list of lowercase booleans
 const booleans = [
   "allowfullscreen",
   "async",
-  "autofocus",
+  "autofocus", // HTMLElement prop
   "autoplay",
   "checked",
   "controls",
   "default",
   "disabled",
   "formnovalidate",
-  "hidden",
+  "hidden", // HTMLElement prop
   "indeterminate",
-  "inert",
+  "inert", // HTMLElement prop
   "ismap",
   "loop",
   "multiple",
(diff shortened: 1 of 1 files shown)


titoBouzout:
Something weird is going on with the tests in GitHub. In local, I see no errors.

packages/dom-expressions/src/constants.js
```diff
@@ -1,16 +1,18 @@
...
+  "alpha", // HTMLInputElement
+  "autofocus", // HTMLElement prop
   "autoplay",
   "checked",
   "controls",
   "default",
   "disabled",
   "formnovalidate",
-  "hidden",
+  // "hidden", // HTMLElement prop - not a boolean
```
ryansolid, line 13: It's not?
Looking at the attribute it looks like it works like a boolean attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden. In that it's presence sets it to true.

Looking at the property: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden
It looks like it takes booleans and or a string but that should still work on property set.

Does empty string mess with it in a weird way? It looks like it is operating fine as is: https://playground.solidjs.com/anonymous/7c281ee5-b35b-45c3-90a0-32c6da8fd7f8. Atleast consistent with how the others work like disabled.

PR #418, opened 2025-05-15
requireImportSource: Don't skip other Babel plugins 

### LINKED ISSUE #417: [babel-plugin-jsx-dom-expressions] requireImportSource skips other Babel plugins
I'm introducing Solid to a project in which I use JSX for other purposes (React, jsx-dom), bundled using Rollup. When I attempt to use the `requireImportSource` option to limit JSX transformation to only my Solid files, I get a Rollup error:

```
[!] (plugin babel) RollupError: [plugin babel] rollup/application.ts: An unexpected situation arose. Please raise an issue at https://github.com/rollup/plugins/issues. Thanks!
```

The cause, from what I can tell, is that `babel-plugin-jsx-dom-expressions` is calling `path.skip()` on the `Program`, which skips not only the current Babel plugin, but other Babel plugins (e.g. `@babel/plugin-transform-typescript`). Instead of using `path.skip()`, I think another mechanism should be used to bypass processing of skipped nodes.

ryb73:
fixes #417. Also adds documentation for `requireImportSource` to the README.

I'm not super familiar with writing Babel plugins, or with the conventions of this project, so any suggestions are welcome.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/README.md
+++ b/packages/babel-plugin-jsx-dom-expressions/README.md
@@ -175,6 +175,29 @@ Removes tags from the template output if they have no closing parents and are th
 
 Removes quotes for html attributes when possible from the template output. This may not work in all browser-like environments the same. The solution has been tested again Chrome/Edge/Firefox/Safari.
 
+### requireImportSource
+
+- Type: `string | false`
+- Default: `false`
+
+When set to a string value, this option restricts JSX transformation to only files that contain a specific JSX import source pragma comment. The plugin will only transform JSX in files that include a comment with `@jsxImportSource` followed by the specified value. If the comment is missing or specifies a different import source, the transformation is skipped for that file.
+
+Example usage:
+```js
+// In babel configuration:
+{
+  plugins: [
+    ["jsx-dom-expressions", {
+      requireImportSource: "r-dom"
+    }]
+  ]
+}
+
+// In your component file:
+/** @jsxImportSource r-dom */
+const template = <div>Hello</div>;
+```
+
 ## Special Binding
 
 ### ref
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/preprocess.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/preprocess.js
@@ -1,6 +1,8 @@
 import * as t from "@babel/types";
 import config from "../config";
 import { isComponent } from "./utils";
+import skipSymbol from "./skipSymbol.js";
+
 const { isValidHTMLNesting } = require("validate-html-nesting");
 
 // From https://github.com/MananTank/babel-plugin-validate-jsx-nesting/blob/main/src/index.js
@@ -25,7 +27,9 @@ const JSXValidator = {
   }
 };
 
-export default (path, { opts }) => {
+export default (path, state) => {
+  const { opts } = state;
+
   const merged = (path.hub.file.metadata.config = Object.assign({}, config, opts));
   const lib = merged.requireImportSource;
   if (lib) {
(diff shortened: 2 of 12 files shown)


packages/babel-plugin-jsx-dom-expressions/README.md
```diff
@@ -1,273 +1,296 @@
-# Babel Plugin JSX DOM Expressions
```
ryb73, line 1: I'm not really sure what happened here... the only changes is additional lines starting at 178

ryb73: Ah that makes sense. Happy to revert the newline changes if that's preferred.

packages/babel-plugin-jsx-dom-expressions/src/shared/preprocess.js
```diff
@@ -25,7 +27,9 @@ const JSXValidator = {
   }
 };
 
-export default (path, { opts }) => {
+export default (path, state) => {
+  const { opts } = state;
+
   const merged = (path.hub.file.metadata.config = Object.assign({}, config, opts));
```
mdynnl, line 33: can be just `state.opts`

packages/babel-plugin-jsx-dom-expressions/src/shared/preprocess.js
```diff
@@ -40,7 +44,7 @@ export default (path, { opts }) => {
       }
     }
     if (!process) {
-      path.skip();
+      state[skipSymbol] = true;
```
mdynnl, line 47: i think `state.skip` should be enough

mdynnl (changes requested):

packages/babel-plugin-jsx-dom-expressions/README.md
```diff
@@ -1,273 +1,296 @@
...
+- Default: `true`
+
+Removes quotes for html attributes when possible from the template output. This may not work in all browser-like environments the same. The solution has been tested again Chrome/Edge/Firefox/Safari.
+
+### requireImportSource
+
+- Type: `string | false`
+- Default: `false`
+
+When set to a string value, this option restricts JSX transformation to only files that contain a specific JSX import source pragma comment. The plugin will only transform JSX in files that include a comment with `@jsxImportSource` followed by the specified value. If the comment is missing or specifies a different import source, the transformation is skipped for that file.
```
mdynnl, line 183: Actually I think we should be checking the exact lib here 

https://github.com/ryansolid/dom-expressions/blob/c9857a7853ccced5551e9b2ad3d7f284b9eb601c/packages/babel-plugin-jsx-dom-expressions/src/shared/preprocess.js#L37

mdynnl: Probably discussed somewhere but we also should be skipping files that `jsxImportSource`-s other lib regardless of `requireImportSource` option which is another thing.

ryb73: > Probably discussed somewhere but we also should be skipping files that `jsxImportSource`-s other lib regardless of `requireImportSource` option which is another thing.

I'm not sure I'm familiar enough with the project to respond meaningfully to this. Do you think that falls within the scope of this PR?

mdynnl: Yeah, not directly related to this PR, which allows other plugins to process when this plugin skips if `jsxImportSource` mismatches, which requires `requireImportSource` option. It's about when the option is not set, then, the plugin would process files regardless of `jsxImportSource` pragma, which I think it should be skipping instead.

For this PR, current changes should suffice 👍.

Issue #429, opened 2025-06-23
Undefined object values in style prop are server rendered as `undefined` string


katywings:
## Current behavior
Undefined values in a object based style prop are server rendered as `undefined`.

## Expected behavior
Style object keys with an undefined value should not be server rendered at all.

## Other infos
Maybe this is a regression? There has been a very similar issue some time ago: solidjs/solid#1224

## Stackblitz / Screenshot
https://stackblitz.com/edit/github-9dhjbayd?file=src%2Froutes%2Findex.tsx

![Image](https://github.com/user-attachments/assets/3d91c264-02e9-4f75-bec1-99cd12e4ec9f)

katywings:
@titoBouzout Whew 🙈, thank you a lot for looking into this ❤️! I am planning to use style tags for css variable declaration a lot more in the future and having to always surround it with a workaround helper is a bit problematic.

titoBouzout:
Yes, working on a patch 🦆

titoBouzout:
@katywings  patch in https://github.com/ryansolid/dom-expressions/pull/430

PR #430, opened 2025-06-30
handle: style property value is undefined in ssr 

titoBouzout:
Fixes https://github.com/ryansolid/dom-expressions/issues/429

The issue is that this 

```js
var _tmpl$ = [
    "<div",
    ' class="bg-(--bg)" style="',
    '">Background should be red</div>',
  ],
...

const Box = (props) => {
  console.log(nope());
  return ssr(_tmpl$, ssrHydrationKey(), "--bg:" + escape(nope(), true));
};
```

ends as 

`<div class="bg-(--bg)" style="-bg:undefined">...`

at first, I fixed it hoisting the value and check for `!= null` but that added a lot of code, and I was not convinced about creating vars because of the gc

```js
const Box = (props) => {
  const _styleprop = escape(nope(), true);
  const _styleprop2 = escape(nope(), true);
  return ssr(_tmpl$, ssrHydrationKey(), (_styleprop != null ? "--bg1:" + _styleprop : "") + (_styleprop2 != null ? ";--bg2:" + _styleprop2 : ""));
};
```

adding a new argument to the escape function was also not a good idea because this `style="--bg:"` still have meaning and breaks.

so then I figured to instead call a function that receives `name` and `value` and only inlines when `value != null`

```js
export function ssrStyleTuple(name, value) {
  return value != null ? name + value : "";
}
...
const Box = (props) => {
  return ssr(_tmpl$, ssrHydrationKey(), ssrStyleTuple("--bg:", escape(nope(), true)));
};
```

Unfortunately, we do not seem to have a test for this exact problem described on the linked issue. and I'm very unsure on how to make one. We have this https://github.com/ryansolid/dom-expressions/blob/main/packages/dom-expressions/test/ssr/ssr.spec.js but changing things there to be undefined still do not reproduce, so honestly idk.  I figured the problem by debugging the compiled code from Katy reproduction, and hacking the solution in node_modules, which it worked so I copied to the repo.

oh I'm not sure if the value being blank changes anything at all in the `ssr(` call, as when value is undefined then `ssr(` will receive an empty string.

Thanks!


katywings:
@titoBouzout `transform-server` only differs from `dom` regarding the settings of `babel-plugin-jsx-dom-expressions`. From what I can tell, this change does and should not affect other preexisting ssr specs, since only the new `ssr.spec.jsx` actually is a `jsx` file, . But I am very happy to be corrected, if I am missing something here 😅.

Issue #343, opened 2024-08-21
Ability to set certain properties / attributes as always static


chiefcll:
Using the universal renderer - I have a property `style` for nodes which is never dynamic. Is there a way to list certain props / attributes as never dynamic to avoid the user needing to add /*@once*/ marker?

ryansolid:
We do a trick where we can evaluate expressions and hoist them out. But that only works within module scope and basically on constants... classList and Style are a bit special cased because their dynamicism is broken down to object properties, but I don't have a generic way to do that. In general I could see the potential of adding a list for saying static I suppose, but we don't have that concept in the DOM.

ryansolid:
fixed in latest 0.40.0

Issue #252, opened 2023-05-20
`/* @once */` seems to work only for first child of object


titoBouzout:
Say we have 

```js
function Comp(props) {
  return (
    <div style={/* @once */ { width: props.width, height: props.height }} />
  );
}
```

generated code is

```js
function Comp(props) {
  return (() => {
    const _el$ = _tmpl$();

    props.width != null ? 
        _el$.style.setProperty("width", props.width) :
         _el$.style.removeProperty("width");

    _$effect(() => props.height != null ? 
        _el$.style.setProperty("height", props.height) :
         _el$.style.removeProperty("height"));

    return _el$;
  })();
}
```

Playground: https://playground.solidjs.com/anonymous/03735897-b261-4420-835c-7e69c27ad978

ryansolid:
Yeah style and classList are special in that they analyze on a per object expression level. So in this case put the once in front of each property.

```js
<div style={ { width: /* @once */props.width, height: /* @once */props.height }} />
```

I realize there is nothing stopping us from oncing the whole object as well.

edemaine:
The current behavior is particularly confusing when `style` is set to something other than an object literal:

```js
<div style={/* @once */ props.style} />
```

Currently `/* @once */` is ignored in this case.  [Playground](https://playground.solidjs.com/anonymous/70270d9f-0a25-4821-9dae-7b062a8d7cae)

In any case, I've [added some documentation](https://github.com/solidjs/solid-docs/commit/9ddbb157510375194374e4f3b96c95974f5488da) about the current behavior so this is less surprising.

titoBouzout:
I was just looking how to fix this, the logic seems to be here https://github.com/ryansolid/dom-expressions/blob/main/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js#L87-L94 
Im staring at this not knowing what to do with that information

titoBouzout:
https://github.com/ryansolid/dom-expressions/pull/258

titoBouzout:
I heard the expectation is to use a memo instead, is the other approach?  Looks to me a bit of a waste, as there's going to be an effect created that is not going to be used, code would be bigger.

For the case in place, can you consider merging it regardless of the future of this feature, it makes it work closer to the intention and of what's expected.

titoBouzout:
see also https://github.com/solidjs/solid/issues/1938

PR #438, opened 2025-07-10
Fix edge cases for `/* @once */` static marker

titoBouzout:
Fixes a couple of edge cases when using `/* @once */`:

1. `@once` now works in nested objects, `<div style={/* @once */ { width: props.width, height: props.height }} />` - Problem was that when the `@once` expression was found, it was removed, so any nested object became dynamic again. 
2. works with object properties as `<div style={/* @once */prop.style}/>` - problem was that `classList` and style were hardcoded to be dynamic, not respecting the `@once` mark.
3. works with spreads `<div { /* @once */ ... props}/>` but only top level - this was just never considered.

A note about spreads, consider the following:

```js
const template28 = (
  <div {/* @once */ ...propsSpread} data-static={/* @once */ color()} data-dynamic={color()} />
);
```

What should happen is that `propsSpread` and `color()` for` data-static` should be static.

If we look at the generated code, it becomes more evident that the issue is `mergeProps` rather than `spread`, as spread has no option but to add the effect for the props that are actually dynamic.

```js
const template28 = (() => {
  var _el$34 = _tmpl$3();
  _$spread(
    _el$34,
    _$mergeProps(propsSpread, {
      get ["data-dynamic"]() {
        return color();
      },
      "data-static": color()
    }),
    false,
    false
  );
  return _el$34;
})();
```

As a compromise, the spread value is now spreaded (when used with `@once`), meaning that `mergeProps` won't track the top level of the object. If access happens in any sub-objects then it will track. This solution is simple enough and possibly covers the immediate expectation.

```js
const template28 = (() => {
  var _el$34 = _tmpl$3();
  _$spread(
    _el$34,
    _$mergeProps({ ...propsSpread }, {
      get ["data-dynamic"]() {
        return color();
      },
      "data-static": color()
    }),
    false,
    false
  );
  return _el$34;
})();
```

Related:
- https://github.com/ryansolid/dom-expressions/issues/252 
- https://github.com/ryansolid/dom-expressions/issues/343
- https://github.com/solidjs/solid/issues/1938
- https://discord.gg/solidjs 
- https://discord.com/channels/722131463138705510/751355413701591120/1391981719128637523 
- https://discord.com/channels/722131463138705510/1361436114484985987/1392936591999373343

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/transform.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/transform.js
@@ -43,7 +43,7 @@ export function transformJSX(path) {
         path.node.leadingComments[0] &&
         path.node.leadingComments[0].value.trim() === config.staticMarker
       ) {
-        path.node.leadingComments = null;
+        path.node.leadingComments.shift();
       }
     }
   });


titoBouzout:
cc  @lxsmnsyc @edemaine

Issue #149, opened 2022-07-10
More ref magic


edemaine:
Inspired by @fabiospampinato's https://github.com/ryansolid/dom-expressions/issues/96#issuecomment-1179727981, here are some ideas for solving some current issues with `ref`:

## Multiple `ref`s on a common element

Currently it's annoying to "merge" two `ref`s together into one `ref` attribute. For example, if you want `props.ref` to be called and to set a local `ref` variable, you need to do something like this:
```jsx
<div ref={(r) => props.ref(ref = r)}/>`
```
This is fairly short but not particularly intuitive or readable. Possible solutions:

1. We could make `ref-1={ref} ref-2={props.ref}` work, i.e., `ref-NUMBER` could work just like `ref`, but they stack like directives do. (Or perhaps `ref:1={ref} ref:2={props.ref}`, but see other suggestion for that syntax below.)
2. We could make `ref={[ref, props.ref]}` work (Fabio's proposal). Only inline arrays would be supported, just like event handlers already have special meaning for inline arrays.

In either case, we'd end up with a `ref` function like this (inlined for Element creation as in `<div/>`):

```jsx
ref(r$) {
  const _ref$ = ref;
  typeof _ref$ === "function" ? _ref$(r$) : ref = r$;
  props.ref(r$);
}
```

## Components that want to provide multiple `ref`s

Sometimes a component wants to expose multiple DOM elements (or other data, e.g. API functionality) to the parent. Currently that needs to be done via callback functions, like so:

```jsx
<Component apiRef={(r) => apiRef = r} viewRef={(r) => viewRef = r}/>
```

Instead, we could imagine a `ref:` prefix that has the same syntactic sugar that `ref` does:

```jsx
<Component ref:api={apiRef} ref:view={viewRef}/>
```

This would compile to:

```js
createComponent(Component, {
  ["ref:api"](r$) {
    const _ref$ = apiRef;
    typeof _ref$ === "function" ? _ref$(r$) : apiRef = r$;
  },
  ["ref:view"](r$) {
    const _ref$ = viewRef;
    typeof _ref$ === "function" ? _ref$(r$) : viewRef = r$;
  },
});
```

So the component could be defined like this:

```jsx
function Component(props) {
  props['ref:api']?.({ ... });
  return <div ref={props['ref:view']}>...</div>;
}
```

[...]

Issue #96, opened 2021-11-10
Feature request: Directives for components


LXSMNSYC:
Currently, the code `<Counter use:example="Hello World" />` compiles into

```js
createComponent(Counter, {
  "use:example": "Hello World"
})
```

We could probably make it work into

```js
createComponent(Counter, {
  ref(node) {
    example(node, 'Hello World');
  },
});
```

since `ref` is already a special prop for both host and components.

What do you think?

ryansolid:
I think the other thing to solve is spreads if we want components to be able to forward these. 

This proposed solution came up during 1.0 rc time period but ultimately wasn't a fan of the inconsistency. Mostly that it puts onus on the component author to forward ref/the right ref. These directives simply may or may not work depending, or not work as expected. Use could only apply to the main ref if multiple were forwarded. Expects precisely a DOM element. There are a lot of unexpected potential. 

Svelte doesn’t support this and Im ok with not as well.

edemaine:
This issue came up again recently in #help, where it was surprising that `<Dynamic component="div">` didn't support directives (or rather, they get assigned as attributes). 

I wonder if it makes sense to handle directives when spreading them into Elements. (This may be what Ryan was suggesting.) Then a component could implement/override/whatever directives if they want, and can pass on other directives via spread.

At first I worried that this wouldn't play well with TypeScript. A component would seemingly need to specify/know every directive they could accept. But explicit enumeration can be avoided: we could say tyat a component can take arbitrary directives using the type `` `use:${string}` `` for keys in `props`. And we could build helper types for this, like DirectiveProps and DirectiveComponent that could get `&`ed on with other types.

LXSMNSYC:
I think part of the issue with `<Dynamic>` is that if you do `<Dynamic component={props.as} use:myDirective />`, how would that be interpreted?

edemaine:
I had the following behavior in mind:
1. If `props.as` is a native element, it would work like a regular directive.
2. If `props.as` is a component, `use:myDirective` (`=true`?) would get passed in as a prop. The component could then do what it wants with the directive, including passing it down via spread to a native element or other component.

This is very close to the current behavior, which just does case 2 in all cases. I'm proposing changing the behavior when you get a native element.

I'm not exactly sure how ergonomic the DX would be here, but this change to `spread` would at least "fix" `Dynamic` when `component` is a native element, which seems like a nice step.

fabiospampinato:
IMO a workable solution for this is the following:

[...]

PR #445, opened 2025-07-12
adds `ref:` namespace

titoBouzout:
Wanted to add `ref` as array, but noticed I could just add a `ref:` namespace and leave everything as is. 

Currently, given how dom-expressions inline code the following just works (albeit is not obvious that it does): 

```js
<div ref={(el)=>el} ref={(el)=>el}/>
```

The problem is with `Dynamic`, this won't work:

```js
<Dynamic component="div" ref={(el)=>el} ref={(el)=>el}/>
```

as compiles to:

```js
  return _$createComponent(Dynamic, {
    component: "div",
    ref: el => el,
    ref: el => el
  });
```

But with the new namespace should work. 

Related to 
- https://github.com/ryansolid/dom-expressions/issues/149
- https://github.com/ryansolid/dom-expressions/issues/96#issuecomment-1179727981

/cc @edemaine

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -547,7 +547,7 @@ function transformAttributes(path, results) {
         (reservedNameSpace ||
           !(t.isStringLiteral(value.expression) || t.isNumericLiteral(value.expression)))
       ) {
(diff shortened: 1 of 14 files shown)


ryansolid:
Yeah this is worth considering. Definitely want to have a decision on this in 2.0. Could add this sooner though without a ton of impact I think.

Issue #390, opened 2024-12-24
optimizing static objects


titoBouzout:
Recording a possible optimization mentioned by Ryan in https://discord.com/channels/722131463138705510/1023298332338167919/1318648799719063693 

this 
```js
<Comp inputProps={{ title: title(), name: name() }} /> 
```
becomes
```js
createComponent(Comp, {
  get inputProps () {
    return { title: title(), name: name() }
  }
})
```
which could be improved to be
```js
createComponent(Comp, {
  inputProps: {
    get title() {
      return title()
    }, 
    get name() {
      return name()
    }
  }
})
```

It would make reactivity more granular I think

lxsmnsyc:
except that's only probably ideal if it's a `Record`. Imagine if it was the user's intent to create a new object instance, rather than be a reactive object.

titoBouzout:
Excellent note, I guess it is somewhat similar to 

```js
<Comp inputProps={title()} /> 
```
which becomes
```js
<Comp inputProps={{get title(){ title() } } /> 
```
if the user wants it static they need to hoist it
```js
const staticc = title() // reading signal 
<Comp inputProps={staticc} /> 
```

PR #446, opened 2025-07-12
[main] avoid getters when possible, and apply getters recursively

titoBouzout:
It adds the optimization you suggested on discord, recorded here https://github.com/ryansolid/dom-expressions/issues/390

It considers spreads and computed keys, such `{[store.key]: "ja"}`, on which case it will keep using a getter.  It also considers `@once`.

It may be possible that we want this optimization in other places 🤔 , as when I search for `"get"` similar code popups around. Can you please have a look? Thanks!

I'm like 95% confident this won't break stuff, but honestly I'm not sure.  

Initially the recursion was a bit messy, I have just simplified it.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/component.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/component.js
@@ -171,15 +171,38 @@ export default function transformComponent(path) {
 
               runningObject.push(t.objectMethod("get", id, [], body, !t.isValidIdentifier(key)));
             } else {
-              runningObject.push(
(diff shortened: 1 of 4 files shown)


lxsmnsyc:
I feel like this is an aggressive optimization. The compiler produces more closure than before.

titoBouzout:
I think the main problem this fix illustrates, is that vars that aren't supposed or expected to be reactive become reactive because of the getter. If the object were inlined when not in need to be reactive, it will behave more "as expected", as in theory you should only expect changes from things that have been defined to change.

PR #447, opened 2025-07-13
wrap member in memo for conditionals 

titoBouzout:
Im really unsure about this one, but in theory it applies this https://github.com/solidjs/solid/issues/2498

/cc @mizulu

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/shared/utils.js
@@ -260,9 +260,10 @@ export function transformCondition(path, inline, deep) {
   if (
     t.isConditionalExpression(expr) &&
     (isDynamic(path.get("consequent"), {
-      checkTags: true
+      checkTags: true,
+      checkMember: true
     }) ||
-      isDynamic(path.get("alternate"), { checkTags: true }))
+      isDynamic(path.get("alternate"), { checkTags: true, checkMember: true }))
   ) {
     dTest = isDynamic(path.get("test"), { checkMember: true });
     if (dTest) {
@@ -287,7 +288,7 @@ export function transformCondition(path, inline, deep) {
       nextPath = nextPath.get("left");
     }
     nextPath.node.operator === "&&" &&
-      isDynamic(nextPath.get("right"), { checkTags: true }) &&
+      isDynamic(nextPath.get("right"), { checkTags: true, checkMember: true }) &&
       (dTest = isDynamic(nextPath.get("left"), {
         checkMember: true
       }));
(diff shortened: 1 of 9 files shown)


mizulu:
@titoBouzout 

to recap on the conversation in discord

The expectations from this feature should be that both a function call (`fn()`) 
and a member access (`obj.prop`) will be treated the same when memo wrapping 
in ternary and `&&` 


### for &&

all these cases should be treated the same in regard to memo and boolean casting 
```
cond() && render()     
cond() && obj.render  
rules.cond && render()
rules.cond && obj.render
```

###  for ternary operator 
all these cases should be treated the same in regard to memo and boolean casting 
```
cond() ? renderA() : renderB()
cond() ? render.a  : render.b
rules.cond ? renderA(): render.b
rules.cond ? render.a : render.b
```


with this change at the minimum you will want to confirm that the tests cover variation 
where member expression and call expression are treated the same in the context of 
conditional rendering, as they are currently treated for call expression (`fn()`)
 



Also interesting to note (current implementation), ternary operator are handled to any depth 
while boolean expressions are not. 

image

titoBouzout:
Thanks @mizulu for the suggestion, copied every conditional test that tests functions to also test member expressions. 

The test is duplicated inline with an `a` added to the end, so you will see output next to each other, input is:

```js
const template16 = <Comp render={state.dynamic || good()} />;
const template16a = <Comp render={state.dynamic || good.good} />;
```

The files are the following,

https://github.com/ryansolid/dom-expressions/blob/a9ddc919be79ee31ca0d3042b4086f566a99017d/packages/babel-plugin-jsx-dom-expressions/test/__dom_fixtures__/conditionalExpressions/output.js
https://github.com/ryansolid/dom-expressions/blob/a9ddc919be79ee31ca0d3042b4086f566a99017d/packages/babel-plugin-jsx-dom-expressions/test/__dom_hydratable_fixtures__/conditionalExpressions/output.js
https://github.com/ryansolid/dom-expressions/blob/a9ddc919be79ee31ca0d3042b4086f566a99017d/packages/babel-plugin-jsx-dom-expressions/test/__dynamic_fixtures__/conditionalExpressions/output.js
https://github.com/ryansolid/dom-expressions/blob/a9ddc919be79ee31ca0d3042b4086f566a99017d/packages/babel-plugin-jsx-dom-expressions/test/__universal_fixtures__/conditionalExpressions/output.js

titoBouzout:
Just checked the output, and it seems like it matches, that's kind of expected because the check its on the same `isDynamic` function.  Added the tests anyway because illustrates nicely when stores are used in place of functions/signals. 

Now, if there's anything else to do/add/change I am very unsure about that, I barely understand the considerations of this output.  @ryansolid if there are some more improvements to be done, let me know, I can add them.

PR #453, opened 2025-08-07
fix regression with inlining of `style` and `classList` 

titoBouzout:
When object is just an identifier it was not adding the effect, such `<div style={style} classList={classes}/>`.
This considers `@once`, for in case It's desired to bail out from the effect.


ryansolid:
Thanks

PR #456, opened 2025-09-05
remove leftover package.json `validate-html-nesting`

tmpluto:
The new version is needed to be able to use `<hr/>` inside of `select` elements.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/package.json
+++ b/packages/babel-plugin-jsx-dom-expressions/package.json
@@ -23,8 +23,7 @@
     "@babel/plugin-syntax-jsx": "^7.18.6",
     "@babel/types": "^7.20.7",
     "html-entities": "2.3.3",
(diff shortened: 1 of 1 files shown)


titoBouzout:
Btw, do you have a repro of the issue? Because Im interested to test if that snippet will work with parse5, I'm tempted to guess you are referring to the new https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/selectedcontent ?  Thanks!

tmpluto:
@titoBouzout  I was using `1.9.5` before. it worked well with `1.9.9`. Thanks.

Removed `validate-html-nesting` from this repo.

titoBouzout:
Great, thanks for testing, and updating the PR. Can you please 
- update the title to "remove leftover package.json `validate-html-nesting`"
- undo the changes to the `pnpm-lock.yaml`

Thanks!

PR #464, opened 2025-10-23
Custom seroval plugins

birkskyum:
For streaming ssr in tanstack start, I've found we need to allow passing in the custom serval plugins that are used with tanstack router. Otherwise there's a mismatch

### DIFF
--- a/packages/dom-expressions/test/ssr/ssr.spec.js
+++ b/packages/dom-expressions/test/ssr/ssr.spec.js
@@ -191,6 +191,18 @@ describe("custom serialization plugins", () => {
     expect(html).toContain("<div>test</div>");
   });
 
+  it("renderToStringAsync accepts plugins option", async () => {
+    const Comp = () => {
+      const pt = new Point(15, 25);
+      sharedConfig.context.serialize("pt", pt);
+      return r.ssr`<div>async</div>`;
+    };
+
+    const html = await r.renderToStringAsync(Comp, { plugins: [PointPlugin] });
+    expect(html).toContain("new Point(15,25)");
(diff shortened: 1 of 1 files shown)


packages/dom-expressions/src/serializer.js
```diff
@@ -19,24 +19,30 @@ const ES2017FLAG =
...
+    HeadersPlugin,
+    ReadableStreamPlugin,
+    RequestPlugin,
+    ResponsePlugin,
+    URLSearchParamsPlugin,
+    URLPlugin,
+  ];
+
+  const allPlugins = customPlugins ? [...customPlugins, ...defaultPlugins] : defaultPlugins
+  console.log('[createSerializer] Total plugins:', allPlugins.length, 'Plugin tags:', allPlugins.map(p => p.tag).join(', '))
```
brenelz: remove these console.logs?

lxsmnsyc:
We need to overhaul the entirety of the serialization API. seroval is currently being used separately by Solid and SolidStart, which brings two issues (version parity and plugin parity). We want dom-expressions to expose a serialization API so that SolidStart doesn't have to independently use seroval on its own and while also keeping the plugin capability the same.

birkskyum:
Right, this ticket isn't about performance, but rather adding flexibility to make some errors go away, so that we can move tanstack start forward. If it opens the door to a shared serializer afterwards, that would be cool, but it's not that important right now.

birkskyum:
since you both have expressed to be ok with this non-breaking tweak, then I'd love to land it, so that we can proceed building out the ssr story for start.

PR #472, opened 2025-11-02
[next] deopt spreads in DOM

titoBouzout:
It deopts spreads, aka spreads should work the same as in JavaScript native.

It fixes:
- being able to override specific props https://github.com/solidjs/solid/issues/2520
- props order is respected https://github.com/solidjs/solid/issues/2532

Notes:
- There may be some possibility for optimization. But I prefer if this works the same as native first, because if not we introduce behaviour that we cant change, (the reason we cannot merge this fix in solid v1). If this works the same as native, then we can see how to optimize it keeping that behaviour in mind.
- Same should be done in SSR
-  `use:` and `ref` arent merged, future of both seem to be undecided.
- It also overwrites duplicated attributes `<div c="1" c="2"/>` discards `c="1"`  producing `<div c="2"/>`

Problems:
- merging `ref` and `use` seems challenging

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -297,6 +297,31 @@ function transformAttributes(path, results) {
     );
     //NOTE: can't be checked at compile time so add to compiled output
     hasHydratableEvent = true;
+  } else {
+    /**
+     * Spreads already de-duplicate attributes.
+     *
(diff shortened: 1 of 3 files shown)


ryansolid:
I want to do this. But I think we should try to keep the DOM and SSR in lockstep. It becomes harder to know the state of things otherwise. `use:` and `ref` are hard. We need make decisions about those on their own probably before deciding how to handle spread for them.

PR #497, opened 2026-04-01
[next] may fix Hydration bug with stateful dom changes

titoBouzout:
Related to https://github.com/ryansolid/dom-expressions/pull/493/changes

In there we were attempting to SSR stuff that wasnt defined in DOM. Issue is that adding attributes/changing childrens, will make SSR have more effects than DOM, making hydration ids drift.


titoBouzout:
Thankfully it seems that this is not needed

PR #498, opened 2026-04-07
[next] align with solid

titoBouzout:
- add missing exports for MathMLElements and friends
- consolidate namespaces into Namespaces


ryansolid:
Thanks

Issue #500, opened 2026-04-12
[SSR] Two regressions in v0.41.0-next.17: `results is not defined` and `JSXEmptyExpression` error


ljho01:
## Bug 1: `results is not defined` in SSR `createElement` when element has both `ref` and spread

**Version:** babel-plugin-jsx-dom-expressions@0.41.0-next.17 (via babel-preset-solid@2.0.0-experimental.16, solid-js@2.0.0-beta.6)

**Reproduction:**
```tsx
function MyComponent(props) {
  let el
  const [local, others] = splitProps(props, ['children'])
  return <div ref={el} {...others}>{local.children}</div>
}
```

Build with `generate: "ssr"` → throws `ReferenceError: results is not defined`

Client build (`generate: "dom"`) works fine.

**Root cause:**
In `createElement` function (index.js:2917), the `ref` attribute handler at line ~2976-2979 references `results.declarations.push(...)`, but `results` is never declared in the `createElement` function scope. 

In v0.40.6, the same `ref` case simply did `return;` without referencing `results`. This appears to be a regression where code was copied from `transformElement$2` (where `results` IS in scope) into `createElement` (where it is not).

## Bug 2: `JSXEmptyExpression` error with bare boolean `muted` attribute on `<video>`

**Version:** Same as above.

**Reproduction:**
```tsx
function MyVideo() {
  return <video src="test.mp4" muted />
}
```

Build with SSR → throws:
```
TypeError: Property right of AssignmentExpression expected node to be of a type ["Expression"] but instead got "JSXEmptyExpression"
```

**Root cause:**
`muted` is listed as a stateful DOM property in the plugin's special handling (index.js:27, `VIDEO: { muted: 1 }`). When a bare boolean attribute (`muted` without `={true}`) hits this code path, the value is `JSXEmptyExpression` which cannot be used as an assignment right-hand side.

**Workaround:** Use explicit `muted={true}` instead of bare `muted`.

## Environment
- solid-js: 2.0.0-beta.6
- babel-preset-solid: 2.0.0-experimental.16
- babel-plugin-jsx-dom-expressions: 0.41.0-next.17
- vite: 8.0.7
- Node: 24

Both bugs only manifest in SSR builds, not client builds.

titoBouzout:
Thanks here are both prs fixing the issues

https://github.com/ryansolid/dom-expressions/pull/501
https://github.com/ryansolid/dom-expressions/pull/502

[...]

PR #501, opened 2026-04-13
[next] default to `true` shorthand attributes on DOMWithState transformation

titoBouzout:
reported in #500 

JSX shorthand attributes like `<video muted />` have a null value node, pr sets it to `true`. 

This is handled in all other places (checked it), Id this but somehow got lost so adds it back


ryansolid:
Thanks

PR #502, opened 2026-04-13
[next] fix SSR createElement ref handling with spreads

titoBouzout:
This has been reported also in  #500 

Think this is a regression introduced in making ref arraays and dropping `use`. 

Honestly, Im like 20% confident on this PR. Sounds plausible refs are dropped in ssr. But Im actually not sure because there are other ref handling in some of the ssr outputs (existing before this pr)

Tests passes, and previous tests outputs have not been modified.

Here are claude explanations, sound plausible but idk (the whole ssr/hydration flow escapes me a little bit)

> SSR createElement (spread path) referenced undefined `results` variable when handling ref attributes, causing a crash. Refs on native elements with spreads are now simply skipped in the compiled output since ssrElement already ignores ref props at runtime.

> Adds MyComponent tests (omit + ref + spread + children) to dom and ssr component fixtures, plus DOM runtime tests for assignment ref, function ref, and signal ref patterns.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/ssr/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/ssr/element.js
@@ -551,13 +551,7 @@ function createElement(path, { topLevel, hydratable }) {
             : node.name.name;
 
         if (hasChildren && key === "children") return;
-        if (key === "ref") {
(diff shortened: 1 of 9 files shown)


ryansolid:
Ok did some edits. But it has been merged.

PR #503, opened 2026-04-13
[next] do not separate dom with state into its own effect

titoBouzout:
We have already removed the check `previousValue != newValue ? input.value = newValue` for any dom with state, as reactivity is the source of truth for these properties(else the dom value won't match the reactive value). 

As we are doing this, then there is no need to have them in a separated effect. 

However, the only concern, is that in these effects there are other things, mostly just short-circuit with `prev != new` which should be fast enough (specially given its only 1 effect). But className/style I'm not sure if short-circuits. In any case, that's seems like a thing to fix for them, and not fault of dom with state.

Separating in a different effect is already being reported as causing problems, because it changes the order of the props.  https://github.com/solidjs/solid/issues/2532#issuecomment-4236026508

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -907,6 +907,8 @@ function transformAttributes(path, results) {
         ) {
           // own effect
           let nextElem = elem;
+          /*
+
           if (isStatefulDOMProperty(tagName, key)) {
             const effectWrapperId = registerImportMethod(path, config.effectWrapper);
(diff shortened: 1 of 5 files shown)


ryansolid:
I think there was 2 reasons for it. But maybe we will realize that again after changing it back here.

titoBouzout:
Would love to hear about the reasons, I'm convinced of my request, but I may be missing something. I feel like previously the split effect has been used as a workaround for things that should have been addressed.

ryansolid:
Cherry picked again to get around merge issues.. This is now merged.

PR #505, opened 2026-04-14
[next] Forbid `select value` (class/classList problem) and ask for `option.selected` instead

titoBouzout:
This PR addresses 2 issues

1. The issue that there are two ways to change what's selected in a `<select>`. One is via `select.value`, and the other is via its options `option.selected`.  (class/classList problem) dom with state will also find this issue when effect re-runs.
2. The other is that `options` can be loaded async, that's why we have the brittle workaround/hack `queueMicrotask`  when its select and value prop. (but Im actually not sure if for solid async a simple queueMicrotask would have worked 100% of the time)

This PR forbids `select.value` similarly how we forbid `<input prop:value={}/>`, and for toggling what's selected you just use the `option`s.  This is what's being suggested in the issues when people have found this difficulty.

1 may be reason enough, but in the case of 2, what worries me the most is that you ship something that works with the queueMicrotask hack, then you make options async and queueMicrotask stops working. So what you ship broke unexpectedly.

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -297,17 +297,6 @@ export function setAttr(path, elem, name, value, { dynamic, prevId, tagName }) {
       t.memberExpression(elem, t.identifier(name)),
       value
     );
-    // handle select/options... TODO: consider other ways in the future
-    // TODO: there may be a race condition here
-    if (name === "value" && tagName === "select") {
-      return t.logicalExpression(
-        "||",
-        t.callExpression(t.identifier("queueMicrotask"), [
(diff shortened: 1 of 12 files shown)


ryansolid:
Yeah.. everyone else supports this. I think this will be a hard sell.

titoBouzout:
This two will beceome obsolete and the problem will just go away
https://github.com/solidjs/solid/issues/1754
https://github.com/solidjs/solid/issues/2241

On the plus point typescript yells at you and theres a message that says to use `<option selected/>` instead, select.value has type never

titoBouzout:
Mmh, yeah think we can add a note on the comment of type itself to warn people that when options are async selecting a value is fragile and options selected should be used instead.

PR #506, opened 2026-04-14
[next] removes mouse events that fire too much from default delegated events , allows opt-in

titoBouzout:
Removes mouse events that fire too much from delegated events. 

Delegated means the event is **attached** to the `document` and **never removed**, so these mouse events are always captured, and the DOM walked up for the existence of the page.

This is a topic we have discussed with @fabiospampinato in the past.

It is of my understanding, that delegated events are nice for when you want to bubble an event from a `Portal` to the component that created the portal. That's cool, however, I think this shouldn't be a default for events that fire too much.

Anyway, I may be right on thinking we can still support the events removed in this PR because we are exporting the `Set` named `DelegatedEvents`, and the function `delegateEvents`. 

So my intention is to remove these events from the defaults, but also make sure we can still opt-in into making these events delegated, for people that desire or need to do so. 

Seems like doing the following should be enough, if my intuition is correct

```js
import { delegateEvents } from '@solidjs/web';
delegateEvents(["mouseover"])
```

### DIFF
--- a/packages/dom-expressions/src/constants.js
+++ b/packages/dom-expressions/src/constants.js
@@ -34,17 +34,10 @@ const DelegatedEvents = /*#__PURE__*/ new Set([
   "keydown",
   "keyup",
   "mousedown",
-  "mousemove",
-  "mouseout",
-  "mouseover",
   "mouseup",
   "pointerdown",
-  "pointermove",
-  "pointerout",
-  "pointerover",
   "pointerup",
   "touchend",
-  "touchmove",
   "touchstart"
 ]);
 


fabiospampinato:
Potentially also worth exploring detaching event handlers for these events instead when nobody else is listening.

Personally I would just delete event delegation for Portal, I don't think it's necessary and it would delete all these problems along with it.

ryansolid:
Hmmm.. torn here.. Original this was classed to basically a certain type of user input event. It's not just portals its also hydration replay.. That being said these events are probably a poor candidate for hydration replay as well. So maybe that's ok.

ryansolid:
Rejecting this. Inferno still delegates the high-frequency mouse/pointer/touch-move events (`onMouseMove` is in their synthetic set — https://www.infernojs.org/docs/guides/event-handling). No prior art for dropping them from the default delegated set.

`DelegatedEvents` / `delegateEvents` remain the opt-in/opt-out if a specific app wants them off.

PR #509, opened 2026-04-16
[next] DOM with state handling in mergeProps, and default* check prev

titoBouzout:
Handles 2 things:

1. Compare defaultValue with previous value from effect, dom property is stateless, but still locked to properties

2. Handle merge props:

At compile time we were ok, at assign time in client we also were ok, but noticed that `mergeProps` had the problem of dom with state having the `prop:` prefix, it would not have merged

`mergeProps` getters need to drop `prop:` in stateful properties to be able to collapse the keys. It doesnt really matter that doesnt have the `prop:` prefix because in client assign time its locked to be a property (the replacement is only done to dom with state properties).

The following:

```js
const propsObj = { value: 1 }
const template97 = <input {...propsObj} value={get()} />
```

was compiling to something on the lines of 

```js
const template97 = (() => {
  var _el$136 = _tmpl$55();
  _$spread(
    _el$136,
    _$mergeProps(propsObj, 
      {
        get ["prop:value"]() {
          return get();
        }
      },
    ),
    false
  );
  return _el$136;
})();
```

it changed now to be the following so it can collapse `value` into 1 thing

```js
const template97 = (() => {
  var _el$136 = _tmpl$55();
  _$spread(
    _el$136,
    _$mergeProps(propsObj, 
      {
        get value() {
          return get();
        }
      },
    ),
    false
  );
  return _el$136;
})();
```

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/element.js
@@ -1293,7 +1293,7 @@ function processSpreads(path, attributes, { elem, hasChildren, wrapConditionals 
       } else {
         runningObject.push(
           t.objectProperty(
-            t.stringLiteral(key),
+            t.stringLiteral(isLockedDOMProperty(tagName, key) ? key.replace(/^prop:/, "") : key),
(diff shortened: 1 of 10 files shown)


ryansolid:
Thanks

PR #512, opened 2026-04-23
[next] Wrap the effect compute argument

### LINKED ISSUE #510: [next?]: effect dependency call directly calls signal source instead of wrapping
# Problem

Suppose we have

```js
<div style={styleSource()} />
```

Solid will compile the style part as:

```js
effect(styleSource, (current, prev) => {
  style(el, current, prev);
});
```

This looks good on paper but suppose that `styleSource` is polymorphic:

```js
const styleSource = (obj) => {
  if (obj) {
    throw new Error('whut');
  }
  return someSource();
};
```

This could cause issues since we unintentionally call `styleSource` with the previous value, far from the initial (and original) intent of the user. An actual example of this should be the atom-style signals (one in `solid-use`: https://github.com/lxsmnsyc/solid-use/blob/00856e01eccfd872f7a42dbfd3ba78ac403129c5/packages/solid-use/src/atom.ts#L12-L16) although of course the user can design the function to take this separately, however, knowing that this is a JS feature, we might as well patch this up.

# Fix

This code should just output `() => styleSource()` directly. I'm not sure if this can happen in other reserved attributes too. Either that or we change `effect` to not pass the previous value to the dependency callback

titoBouzout:
fixes #510
 
Wrap the effect compute argument in an arrow when the JSX has a single dynamic attribute, so the source expression isn't called with the previous value as its first  argument

### DIFF
--- a/packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
+++ b/packages/babel-plugin-jsx-dom-expressions/src/dom/template.js
@@ -4,9 +4,9 @@ import {
   getConfig,
   getNumberedId,
   getRendererConfig,
-  inlineCallExpression,
   isStatefulDOMProperty,
-  registerImportMethod
+  registerImportMethod,
+  wrapForEffect
 } from "../shared/utils";
 import { setAttr } from "./element";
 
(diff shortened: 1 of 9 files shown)


ryansolid:
@changeset-bot recheck

PR #515, opened 2026-04-29
fix document root reactivity

brenelz:
## Summary
Fixes document-root rendering/hydration so lazy memo-owned content remains reactive after `render(..., document)` / `hydrate(..., document)`.
## Problem
The `element === document` render path only called `flatten(code)`. With `@solidjs/signals@2.0.0-beta.9`, lazy memos can auto-dispose when read without an active subscriber. That means a top-level lazy memo tree can hydrate initially, then stop responding to later signal updates.
This showed up in Tanstack Solid Start-style full-document hydration, where navigation state updated but nested rendered content could stay stale.
## Fix
For document-root rendering, evaluate the app tree once and keep its flattened output observed in a render effect:
```js
const tree = code()
effect(() => flatten(tree), () => {})
```
This preserves the existing document-root behavior of not inserting into document, while ensuring the root JSX tree remains subscribed and reactive.
Tests
- Upgraded the test dependency to @solidjs/signals@2.0.0-beta.9 to reproduce the auto-dispose behavior.
- Updated the test adapter for beta9 createMemo / createRenderEffect signatures.
- Added a regression test covering lazy memo-owned content rendered with document as the root.
Verified with:
pnpm --filter dom-expressions test -- --selectProjects browser test/dom/render.spec.jsx --runInBand

### DIFF
--- a/packages/dom-expressions/src/client.js
+++ b/packages/dom-expressions/src/client.js
@@ -55,7 +55,11 @@ export function render(code, element, init, options = {}) {
     dispose => {
       disposer = dispose;
       if (element === document) {
-        flatten(code);
+        const tree = code();
+        effect(
+          () => flatten(tree),
+          () => {}
+        );
       } else {
         const tree = code();
         insert(
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@rollup/plugin-commonjs": "24.0.0",
     "@rollup/plugin-json": "6.0.0",
     "@rollup/plugin-node-resolve": "15.0.1",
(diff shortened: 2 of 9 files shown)


packages/dom-expressions/src/client.js
```diff
@@ -55,7 +55,11 @@ export function render(code, element, init, options = {}) {
     dispose => {
       disposer = dispose;
       if (element === document) {
-        flatten(code);
+        const tree = code();
```
brenelz, line 58: This is the main change

ryansolid:
Thank you

PR #518, opened 2026-05-09
fix class reconciliation bug

mzdunek93:
Example of the bug [here](https://stackblitz.com/edit/github-gj63q1tt-y8io2gkm?file=src%2Fapp.css,vite.config.ts,src%2FApp.tsx) (turn on both of the colors and then switch off one of them, the text will be dark instead of white).

This is a proposed solution, might use some refactoring.


ryansolid:
Hey I went with another fix. It should be in the next release. I did use the test it was helpful. Thank you.

PR #522, opened 2026-05-14
1 Template per top level element

danielrkling:
Fixed bug where expressions at top level did not cleanup correctly. Solution was switching to 1 template for each top level element (Matches JSX compiler behavior) instead of 1 template for the whole tree. 

Also added position info in the errors.


ryansolid:
Thank you

Issue #450, opened 2025-07-28
Feature request: Add support for the `style:` prefix


AFatNiBBa:
I noticed [this](https://github.com/ryansolid/dom-expressions/blob/8c822749acf1bb20b000b0943ca1f0f8e28199d2/packages/dom-expressions/src/jsx-h.d.ts#L1195) line, do you intend to add this feature?
I can think at least 2 benefits from adding this just from the top of my head:
1. Allow for fine-grained reactivity on the props (Check [this](https://playground.solidjs.com/anonymous/987d688f-aa7d-452e-9636-825eaea32e37) playground)
2. Avoid having to wrap props with hyphens, as JSX supports them out of the box
    ```tsx
    <div style={{ display: "grid", "grid-template-columns": "max-content 1fr" }} />
    ```
    Becomes simply
    ```tsx
    <div style:display="grid" style:grid-template-columns="max-content 1fr" />
    ```

AFatNiBBa:
Interesting.
Are the effects merged to meet the expectations about what should happen when an object is used?
Or is there some other reason to use a single effect?

titoBouzout:
Is just to avoid creating too many effects, think is a performance optimization.  Everything is cached in local vars and compared with the previous value. The assumption is that props are deterministic, and only change because some tracked value changed. That means that functions that aren't deterministic should be memoized, and non-deterministic objects should be hoisted, to avoid values unexpectedly changed.

As for the original question of supporting `style:background={red}`  I will let Ryan reply. personally I like that form.

AFatNiBBa:
In the unlikely event that this went unnoticed, `spread()` doesn't currently handle this syntax (Check [this](https://playground.solidjs.com/anonymous/adbda027-eba1-45c2-9690-77d371369170) playground).

As for the merged effect, I was just thinking about the off chance that the computations were performance-intensive: Using the object syntax I expect all of them to change each time so I will memoize what's needed, but I wouldn't expect this new syntax to behave in the same way (I'm probably just worrying too much though)

titoBouzout:
PR for supporting "style:" namespace in spreads is in https://github.com/ryansolid/dom-expressions/pull/462

AFatNiBBa:
Since it was mentioned, just to get a bit of context, what's the current state of the class/classList argument as per Solid 2.0?

AFatNiBBa:
Wasn't the problem that `class` and `classList` would override eachother when updating?
[...]

PR #523, opened 2026-05-15
[next] Removes `style:`, `class:` namespaces, `@once` feature  

titoBouzout:
Removals:

- removes support for `style:` namespace, some discusion https://github.com/ryansolid/dom-expressions/issues/450
- removes `class:` namespace - class/classList problem (two values refering to the same overwrite each other reactively and wont merge properly in spreads)
- removes `@once` static helper.   somewhat realted https://github.com/solidjs/solid/pull/2688 and in any case people can use refs I suppose
- fixes types, something went wrong on "feat: remove on namespace events" commit and it removed more types from JSX namesapce than needed https://github.com/ryansolid/dom-expressions/commit/d5cd499af8b8c2f07d86402561ae1238651f873c
- add `Date` to autodiscovered properties for `valueAsDate`


ryansolid:
I don't know about removing @once.. maybe it should be renamed.. but there is a precedent for like.. the compiler got this wrong it is static.

titoBouzout:
I think making static things as described in the document https://github.com/solidjs/solid/pull/2704/changes should be more than enough and works consistently, thing that `@once` doesnt. 

The nature of everything is reactive and obscured, comes from the getters, is not tecnically the compiler wrong. Having `@once` doesnt really help. The problem is merged effects + getters + once not doing what it should. 

I think having a way to do 1 thing and from that thing to work cosistently is better.

ryansolid:
I think for values that can be reactive the docs recommendation is correct. However.. just because an object has a property or a function is called in the JSX doesn't mean those are reactive values. The heuristic can be wrong and we need to provide escape hatches. As I said `@once` is probably the wrong name since it suggests reactivity.. Maybe it should just be `@static`

titoBouzout:
I know all of that. `@once` is still broken, it doesn't work with objects and objects that are inlined like `size={props.size}`, and we recently learned that it changes order of props. 

The structure clone of the documentation (besides using snapshot for stores) is better because works 100% of the time and is idiomatic.  Not sure why are you so resilient  to drop a feature that barely works.

PR #530, opened 2026-06-17
Add types from vscode

huseeiin:
added descriptions for elements and events from vscode


titoBouzout:
Can you please keep both links as it was before? One link is for authoring html (the attributes) and the other link is for properties + methods (for using the element in refs for example)

titoBouzout:
Also, notice the comment at the top of the file, 
https://github.com/ryansolid/dom-expressions/blob/062d23cc29731e8c2281ddfa36188d438a90e21f/packages/dom-expressions/src/jsx.d.ts#L1-L4

ryansolid:
Not taking VS Code element/event JSDoc onto the 1.x `jsx.d.ts`. Types for 2.x live in the Solid repo.

PR #531, opened 2026-06-24
Research/oxc ast native

ryansolid:
## Summary
- Adds the experimental `jsx-dom-expressions-compiler` package with an Oxc/N-API AST-native transform API.
- Covers DOM, hydratable DOM, SSR, hydratable SSR, universal, dynamic, wrapperless, and no-inline-styles fixture paths.
- Mirrors Babel plugin behavior across supported options and adds a changeset for the new compiler package.

## Test plan
- `pnpm test`
- `pnpm --filter jsx-dom-expressions-compiler test`
- `pnpm --filter jsx-dom-expressions-compiler lint`
- `cargo fmt --manifest-path packages/jsx-dom-expressions-compiler/Cargo.toml --check`


packages/jsx-dom-expressions-compiler/src/dom/attrs.rs
```diff
@@ -0,0 +1,566 @@
...
+            operations.push(self.spread_attribute_statement(
+                attributes,
+                element_id,
+                skip_children,
+            )?);
+            return Ok(());
+        }
+
+        for attr in attributes {
+            if let Some(statement) = self.no_inline_style_attribute_statement(attr, element_id)? {
```
atk: That inline loop has a lot of code duplication, maybe consider joining conditions with Option.or and `||`:

```ts
if let Some(statement) = self.no_inline_style_attribute_statement(attr, element_id)
  .or(self.prop_attribute_statement(attr, element_id)
  .or(self.child_property_attribute_statement(attr, element_id)? {
  operation.push(statement);
  continue;
}
if self.class_array_attribute_operations(attr, element_id, template, operations)? ||
  self.style_object_attribute_operations(attr, element_id, template, operations)? {
  continue;
}
match ...
```

ryansolid: Thanks, cleaned this up in 3790d0a3 by extracting the optional statement-producing checks into `attribute_operation_statement` and short-circuiting class/style template operations.

PR #533, opened 2026-07-04
fix(client): don't let transient detached values clobber tracked nodes during hydration

brenelz:
Fixes bug 1 of solidjs/solid#2801 — a bare async value directly inside `<Loading>` next to element siblings (`<Loading fallback={...}><button>Refresh</button>{data()}</Loading>`) hydrates fine, but the first `refresh()` leaves the old value behind as a ghost (`Refreshvalue-1value-2`). Wrapping the value in a `<div>` avoids it, which pointed at positional text adoption rather than element claiming.

## Root cause

During hydration a boundary briefly renders its fallback — a freshly created, **detached** node — before resuming with the claimed children. In `insert()`, `insertExpression` correctly leaves the server-rendered DOM untouched while hydrating, but `current = value` still ran unconditionally, so the tracked server nodes (`[button, #text "value-1"]`) were replaced by the never-inserted fallback. On the resume pass, `normalize()` then looked at `current[1]` to adopt the existing text node, found the fallback div instead, and created a *detached* replacement text node. The first post-hydration update reconciled against that detached node — removal was a no-op (`parentNode === null`) — and the real server text node stayed in the DOM untracked.

## Fix

A value may only become the tracked `current` during hydration if all its DOM nodes are connected, i.e. it re-identifies claimed/adopted server nodes. Transient values holding detached nodes (the fallback pass) no longer clobber tracking, so the resume pass adopts the real text node and post-hydration updates reconcile against the DOM that's actually there.

Non-node values (strings, `undefined`) pass through unchanged: rejecting `undefined` too would regress the spread + `innerHTML` case (solidjs/solid#2737), where `props.children` is legitimately `undefined` while the content belongs to `innerHTML` — keeping stale node tracking there makes the next update wipe the content via `cleanChildren`.

## Tests

- New regression test in `test/ssr/hydrate.spec.js` modeling the exact sequence with runtime primitives (insert with server child nodes → detached value while hydrating → claimed element + adopted text → post-hydration swaps). Fails on current `next` with `value-1value-2`; passes with the fix.
- Full runtime suite: 473/473. hyperscript 69/69, tagged-jsx 179/179. (`babel-plugin-jsx` has 2 pre-existing `dynamic.spec.js` failures on pristine `next` tip, unrelated; `jsx-compiler`'s Cargo.lock doesn't parse with my local Rust toolchain so it wasn't run.)
- Also validated end-to-end in the solid repo (with this file patched into the installed runtime): the solidjs/solid#2801 bug-1 repro flips to passing, its div-wrapped control stays green, and solid's server (115), client (299), and hydration suites pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/runtime/src/client.js
+++ b/packages/runtime/src/client.js
@@ -348,7 +348,7 @@ export function insert(parent, accessor, marker, initial, options) {
         () => normalize(value, current, multi),
         inner => {
           insertExpression(parent, inner, current, marker);
-          current = inner;
+          if (!isHydrating(parent) || tracksDOM(inner)) current = inner;
           host && tagHost(current, host);
         },
         prev !== undefined && !(options && options.schedule)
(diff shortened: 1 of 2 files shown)


ryansolid:
Thanks for the careful diagnosis — the fallback-pass-clobbers-`current` sequence was real, and modeling it with raw runtime primitives made it easy to verify independently. Two things happened since you opened this: the hydration rework that landed on `next` (hole id scopes + the `$df` region re-claim) changed how boundary fallbacks interact with tracked nodes during hydration, and with that in place your regression test passes without the `tracksDOM` gate — I applied just the test from this PR against current `next` and it's green as-is, along with the original solidjs/solid#2801 bug-1 repro.

So I'm declining the runtime change as superseded, but I've adopted your test into `next` with a `Co-authored-by` credit so this exact sequence stays pinned against regression. I'd also add that gating what `current` may track based on `isConnected` made me nervous as a general mechanism — it special-cases detached values in a hot path to compensate for a boundary behavior that was itself the bug — so I'm glad the root-cause fix made it unnecessary.

Appreciate the contribution — the test is shipping in the next beta.

PR #535, opened 2026-07-06
fix: identify slots by per-insert tokens so adjacent slots can exchange nodes

yumemi-thomas:
Fixes solidjs/solid#2830 — nodes migrating between **adjacent** JSX expression slots are destroyed by the slot they left. This closes the remaining gap in the slot-ownership model this repo shipped for solidjs/solid#2030 ("Support moving DOM Element references inside JSX"): migration works across parents and across separated slots, but adjacent slots silently eat nodes. The everyday shape is two `<Show>`s trading kept-alive panels on a layout flip — each toggle destroys one panel. The full failure class, each pinned by a test:

- exchange between adjacent slots drops a node per swap (`"12"` → `"1"`), in both the tail-position and shared-static-marker geometries
- a single node migrating to the adjacent slot **vanishes entirely**, in both directions
- arrays exchanging members across adjacent slots throw `NotFoundError` mid-reconcile
- a rotation across three slots destroys two of the three elements
- an interleaved text expression does not separate the slots
- related, present even without migration: a null-marker slot emptied via `[]` loses its anchor, so refilled content appends after later slots' content
- **post-hydration the same JSX works** (hydratable output emits a marker pair per slot), so this is also a CSR/SSR behavioral divergence

Scope note: every move here happens through JSX expression values — this is the "Solid controls both sides" case #2030 shipped, not the externally-moved-nodes class declined in solidjs/solid#2515. External-mutation behavior is unchanged.

## Root cause

The compiler gives adjacent expression slots the **same insertion marker** (`null` at the tail, or the next static element — verified for raw expressions and components alike), and the ownership predicate uses the marker as identity:

```js
const owns = el.parentNode === parent && (!tag || tag === marker);
```

Tail slots never tag at all (`if (marker)`), so both slots consider every node fair game; marker-sharing slots tag with the same node, so `tag === marker` is true across slots. Either way two logical slots collapse into one ownership region, and one slot's cleanup removes the node its neighbor just claimed.

## Fix

Ownership is keyed on a per-`insert()` slot identity token; the marker keeps its positioning role. Gated so the hot path is untouched:

- **Single-slot parents (the For-list shape) run the exact previous instruction stream** — tokens engage only when a parent registers a second slot (`parent._$slots`, append-only, creation order).
- Ownership checks accept **token-or-marker**, so content inserted before a parent becomes multi-slot stays owned across the transition; adoption always retags, which is what keeps the permissive untagged case safe.
- A slot whose entire content migrated away in the same flush it receives fresh content recovers its position from the next live slot's region — this also fixes the pre-existing `[]`-refill mis-order.
- Reconcile's `after` anchor gains a guard for the migration case where `tail.nextSibling` is itself an incoming node.
- One-shot inserts stay token-less (direct callers legitimately thread `current` across separate calls) but register for positional queries.

Registry entries live as long as their parent element and hold small slot records plus each slot's last value array; eager release on disposal is deliberately **not** in this PR — it only benefits manual `insert()` reuse against long-lived parents, and is staged separately with its own tests if wanted. DOM renderer only; `universal.js` intentionally unchanged (older host environments, expando/proxy interactions, and these migration patterns aren't idiomatic there).

## Tests

- 27 new tests (25 in `test/dom/slot-ownership.spec.jsx`, 2 hydration in `test/ssr/hydrate-slot-ownership.spec.js`): the failure shapes above, both directions, function-valued slots (the nested-effect path components compile to), dynamic late slot registration, three-parent migration chains, mixed node→text→node cycles, same-node-in-two-slots (last insert wins, no crash), a registry-growth stress documenting the append-only tradeoff at 100 stale entries, and controls pinning the working boundary (separated slots, cross-parent, post-hydration).
- Full runtime suite: 497/497 across browser/server/universal projects.
- Validated end-to-end in the solid repo with this runtime patched into the installed package: the solidjs/solid#2830 repro (6 tests) flips grepasses with zero regressions.

## Performance

Benchmarked old-vs-new through built dists in isolated headless-Chrome processes, 9 loads per variant, medians with distribution overlap comput

| tier | result |
|---|---|
| js-framework-benchmark keyed ops (9 ops) | 9/9 overlapping distributions, max median delta +3.0% |
| realistic app (tabbed dashboard, 200-row list, kept-alive panel swap) | mount identical (0.567ms both); interactions overlap |
| targeted runtime workloads (reverse/swap/replace/create ×2/shared updates) | 6/6 overlapping |

Across all 20 measured operations, the **only statistically disjoint row is the kept-alive tab switch — where the old runtime destroys a panel  absence of work. Residual cost is ~0.02µs/update confined to multi-slot parents. The swap1k-guarded branch split in `reconcileArrays` ispreserved.

Alternative considered: compiler-emitted per-slot markers (hydratable output already produces them, which is why hydrated pages don't have thiswith zero runtime bookkeeping but permanent extra template/DOM nodes at every adjacent-expression site and compiler/runtime lockstep. The runtime token approach keeps DOM shape and works with existing compiled output; happy to pivot if you prefer the marker route.

Full disclosure: I wrote this with the help of an AI assistant (Claude Fable 5) and reviewed every line before pushing. All new tests were written first and confirmed failing on the unfixed code

### DIFF
--- a/.changeset/slot-token-ownership.md
+++ b/.changeset/slot-token-ownership.md
@@ -1,0 +1,30 @@
+---
+"@dom-expressions/runtime": patch
+---
+
+Fix adjacent expression slots destroying each other's nodes when content
+migrates between them. Adjacent compiled slots can share a physical
+insertion marker (`null` at the tail, or the next static element), so
+marker-based `$$SLOT` ownership collapsed neighboring slots into one
+region: exchanges dropped nodes, single-node migrations vanished in both
+directions, and array exchanges could crash `reconcileArrays` mid-update —
+while the same JSX worked after hydration, where hydratable output emits a
+dedicated marker pair per slot.
+
+Ownership is now keyed on per-`insert()` slot identity. Single-slot
+parents keep using the physical marker as the ownership tag (their
+instruction stream is unchanged); parents hosting multiple expression
+slots switch to per-slot identity tokens registered on `parent._$slots`.
+Ownership checks accept token-or-marker so content inserted before a
+parent becomes multi-slot stays owned across the transition. When a slot's
+entire content migrates away in the same flush it receives fresh content,
+its position is recovered from the next live slot's region — which also
+fixes a pre-existing mis-order where a null-marker slot emptied via `[]`
+appended refilled content at the parent's end, after later slots' content.
+
+Scope: DOM renderer (`client.js`) only. `universal.js` is intentionally
+unchanged — universal hosts target older JS environments, expando writes
+on platform nodes can collide with proxy-based node wrappers, and the
+JSX-DOM migration patterns this fix addresses are not idiomatic on
+non-DOM platforms. If a real case surfaces on a universal renderer it can
+be revisited with a host-appropriate storage strategy.
--- a/packages/runtime/src/client.js
+++ b/packages/runtime/src/client.js
@@ -12,6 +12,7 @@ import {
   flatten
 } from "rxcore";
 import reconcileArrays from "./reconcile";
+import { createSlot, registerSlot, slotBoundary } from "./slots";
 import { DOMWithState } from "./constants";
 export {
   DOMWithState,
@@ -313,6 +314,8 @@ export function ref(fn, element) {
 
 export function insert(parent, accessor, marker, initial, options) {
   const multi = marker !== undefined;
+  // Adjacent slots can share a physical marker; token identity separates them.
+  const slot = multi ? createSlot(initial) : undefined;
   const host = options && options.host;
   if (multi && !initial) initial = [];
   if (isHydrating(parent)) {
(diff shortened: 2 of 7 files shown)


ryansolid:
Thank you for this — the diagnosis here is excellent and exactly right. The marker doing double duty as the `$$SLOT` ownership identity is precisely how two adjacent slots collapse into one ownership region, and your failure-class breakdown (tail vs shared-static geometries, the `[]`-refill mis-order, the CSR/post-hydration divergence) saved us a lot of investigation time. The test corpus and the benchmark rigor are genuinely appreciated.

That said, I'm going to take you up on the pivot you offered at the end rather than merge the token approach. Zero runtime bookkeeping was a big part of why I was willing to adopt migrating nodes at all, and even carefully gated, the token route adds a registry, a second identity check in the ownership predicate, and per-parent state that lives as long as the element — costs that smear across the system to serve one compile-time-knowable case. The compiler already knows exactly which parents host multiple slots, and hydratable output has always emitted per-slot markers for them (which is why these shapes work after hydration, as you noted). So the fix that just landed on `next` makes CSR output match that same geometry: every slot in a multi-slot parent gets its own truthy marker — riding the immediately following static sibling when one exists, otherwise a dedicated `<!>` placeholder. Single-slot parents (the hot list-container shape) compile byte-identically to before, the runtime is untouched, and CSR/SSR behavior converges by construction. The trade is a comment node per adjacent-expression site, which we've accepted as the honest cost of the feature.

Your failure shapes are pinned as runtime tests against real compiled JSX (exchange, both migration directions, array member exchange, three-slot rotation, interleaved text, the one-sided ghost case, `[]`-refill ordering, plus the working-boundary controls) — all verified failing against the old compiler output. Ported to both the Babel plugin and the Rust jsx-compiler. It will be in the next beta release, and solidjs/solid#2830 closes with it.

Closing in favor of the compiler-marker implementation, but this PR materially shaped it. Thanks again for the care you put into this.

PR #536, opened 2026-07-06
fix(client): delegated events reach outer roots across nested render roots

yumemi-thomas:
Fixes solidjs/solid#2832 — when a render root is mounted inside another root's DOM (an embedded widget or microfrontend calling `render()` into a slot of the host app), the outer root's **delegated** handlers never fire for events from the inner root, even though the native event bubbles through the outer root's elements: a plain `addEventListener` on the same wrapper fires while the delegated `onClick` on it doesn't. The everyday casualty is any delegated handler above an embedded root — click-outside-to-dismiss, analytics capture, modal scrims — silently dead inside the embedded region, with native listeners still working, which makes it maddening to debug. This contradicts the natural reading of the root-owned delegation docs (*"`event.stopPropagation()` inside a nested Solid root can prevent outer roots … from observing the native event"* — implying un-stopped events are observed), and it's a regression from 1.x, where document-level delegation walked the composed path across roots.

The same root-scoping affects **hydration replay**: `runHydrationEvents` picked only the nearest registered root, so a click queued during hydration would handle just the innermost segment even once live dispatch is fixed.

## Root cause

Every registered container receives the bubbling native event, innermost first. The first listener marks it consumed for all others:

```js
if (e[$$EVENT_OWNER]) return;   // ← outer root's listener bails out here
...
e[$$EVENT_OWNER] = owner || true;
```

…and its walk (correctly) stops at its own root boundary. So the inner root never invokes outer handlers, and the marker then prevents the outer root's listener from walking its own segment.

## Fix

`$$EVENT_OWNER` becomes "the boundary of the most recent walk" instead of a consumed flag. An ancestor container whose subtree contains the recorded boundary resumes the walk *from that boundary element* up to its own boundary; unrelated containers still bail:

```js
if (prev === true || prev === container || !container.contains(prev)) return;
resumeNode = prev;
```

Resuming *at* the boundary is load-bearing: walks fire a boundary's own handlers only when it is the event target, so in the normal case the inner root's container element (outer-owned JSX that may carry its own handlers) hasn't fired yet. In the one case where it has — the click landed directly on the container — the resume steps to the next node up instead, and skips entirely if that lands on the outer root's own boundary. Each root's handlers run exactly once, innermost-out, matching native bubbling order. `stopPropagation()` composes for free: it stops the native event before outer containers' listeners run. `runHydrationEvents` now relays queued events through all matching roots innermost-first, so pre- and post-hydration clicks behave identically.

Behavior change to be aware of: code using nested roots to *sandbox* clicks from outer handlers — i.e. relying on this bug — will see outer handlers fire again; `stopPropagation()` remains the documented isolation mechanism. Known limitation, unchanged from before: portal content mounted outside all roots and owned by a *nested* root still can't reach outer roots (the native event never touches their containers). Non-nested apps are unaffected by construction — the resume path is unreachable unless an inner root already recorded a boundary, so there is no cost on the common path.

## Tests

- 11 new tests in `test/dom/nested-roots.spec.jsx`: the basic case with a native-listener control, three-level nesting firing innermost-out once each, sibling nested roots staying isolated, `stopPropagation` suppression across roots, between-root handlers with correct `currentTarget`, the nested root's container element firing exactly once — including when the click lands directly on the container, and when that container is a direct child of the outer root's boundary — bound-handler data (`onClick={[fn, data]}`) on the resumed walk, hydration replay across nested roots, and a nested root inside an outer-owned external portal (wired the way `Portal` actually registers: `_$host` + `registerDelegatedContainer`).
- Full runtime suite: 483/483 across browser/server/universal — all existing events, shadow-DOM, portal, and replay tests unaffected.
- Validated end-to-end in the solid repo with this runtime patched into the installed package: the solidjs/solid#2832 repro flips green, and both solid-web's client suite (309 passing) and its separately-configured hydration suite (byte-identical failure set to pristine baseline) show zero regressions attributable to this change.

### DIFF
--- a/.changeset/nested-root-delegation.md
+++ b/.changeset/nested-root-delegation.md
@@ -1,0 +1,24 @@
+---
+"@dom-expressions/runtime": patch
+---
+
+Fix delegated events never reaching outer roots when a render root is
+rendered inside another root's DOM (embedded widgets, microfrontends).
+The first (innermost) container listener marked the event consumed for
+every other root, so an outer root's delegated handlers were silently
+skipped even though the native event bubbled through its elements: a
+plain `addEventListener` on the same element fired while the delegated
+handler didn't.
+
+`$$EVENT_OWNER` now records the boundary of the most recent walk instead
+of a consumed flag: an ancestor container whose subtree contains that boundary
+resumes the handler walk from it up to its own boundary, so each root's
+handlers fire exactly once, innermost-out, matching native bubbling.
+`stopPropagation()` inside a nested root still suppresses outer roots (it
+stops the native event before their listeners run), and hydration event
+replay now relays queued events through all matching roots innermost-first
+so pre- and post-hydration clicks behave identically. Apps that relied on
+nested roots to isolate clicks from outer handlers should use
+`stopPropagation()`, which remains the documented mechanism. Non-nested
+apps are unaffected; the resume path is unreachable unless an inner root
+already handled the event.
--- a/packages/runtime/src/client.js
+++ b/packages/runtime/src/client.js
@@ -597,14 +597,34 @@ export function runHydrationEvents() {
         const [el, e] = events[0];
         if (!completed.has(el)) return;
         events.shift();
-        let match;
+        let matchContainer, matchState, matchDistance, matches;
         for (const [container, state] of delegatedContainers) {
           if (!state.handlers.has(e.type)) continue;
           const entry = findOwner(e.target, state);
-          if (entry && (!match || entry.distance < match.distance))
-            match = { container, state, distance: entry.distance };
+          if (!entry) continue;
+          if (matchContainer) {
+            if (!matches)
+              matches = [
+                {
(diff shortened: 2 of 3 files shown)


ryansolid:
This one's a merge — great catch and a really clean resume mechanism. Verified the repro against `next` (outer delegated handler silently dead while a native listener on the same element fires), reviewed the boundary-resume semantics including the corner where the click lands directly on the nested root's container, and ran it through both repos: full dom-expressions suites and the whole solid monorepo including the hydration parity harness, all green. The innermost-out ordering matching native bubbling and `stopPropagation()` composing for free is exactly the behavior the root-owned delegation docs promise.

Cherry-picked onto `next` with your authorship preserved; it ships in the next beta alongside the adjacent-slot fix. Closing since the branch history is maintained locally ahead of the release push — the commit will show up on `next` shortly. Thanks for another excellent contribution.

PR #540, opened 2026-07-09
fix(compiler): scope-wrap dynamic children holes in the ssr spread path

brenelz:
## Summary

The hole-owner-id-scopes change added the ssr generate's `scope()` wrap in `transformChildren` (the template path) but never in `createElement`, the spread-element path that renders through `ssrElement`. The dom generate scope-wraps the matching insert accessor regardless of spread, so a shape like `<a {...props}>{children()}</a>` reserved a hydration id slot on the client but not on the server — every hydration key allocated after the hole drifted and following siblings went unclaimed (duplicated DOM, "unclaimed server-rendered node" warnings).

Hit in practice upgrading TanStack Router to Solid 2.0.0-beta.16, where every `<Link>` desynced the ids of all its following siblings.

## Fix

`createElement` now applies the same wrap as `transformChildren`: dynamic children holes that can allocate hydration ids evaluate under their own owner scope.

## Testing

Regression-tested end-to-end in the runtime hydration suite; the new test (`packages/runtime/test/ssr/hydrate-spread-scope.spec.js`) fails with the exact sibling-unclaimed symptom when the wrap is removed.

### DIFF
--- a/.changeset/fix-ssr-spread-child-scope-wrap.md
+++ b/.changeset/fix-ssr-spread-child-scope-wrap.md
@@ -1,0 +1,5 @@
+---
+"@dom-expressions/babel-plugin-jsx": patch
+---
+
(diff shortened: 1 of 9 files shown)


brenelz:
Might need a rust compiler change too?

PR #551, opened 2026-07-24
feat(compiler): expose a host-independent Rust core

yumemi-thomas:
## Summary

Expose the JSX compiler as a host-independent Rust API that can be used without linking the Node/N-API adapter.

This separates the compiler into three layers:

- a public Rust compiler core
- a thin Node/N-API adapter
- shared lowering internals using host-independent error types

The existing Node `transform()` API remains enabled by default and preserves its current behavior.

## Changes

- Add the public `compile()` Rust API with owned options, output, and errors.
- Build the crate as both `rlib` and `cdylib`.
- Make N-API dependencies optional behind the default `node` feature.
- Move the existing Node interface into a dedicated adapter.
- Remove N-API `Result` and `Error` types from shared lowering internals.
- Add a targeted `--no-default-features` integration test.
- Add a patch changeset for `@dom-expressions/compiler`.

Most of the apparent diff is the existing `lib.rs` implementation being separated into `compiler.rs` and `node_adapter.rs`. The lowering files only contain the mechanical error-type decoupling needed to compile without N-API.

## Motivation

Rust consumers such as `solid-checker` need to invoke the compiler directly without depending on a Node host or N-API runtime.

This PR establishes that host-independent boundary. Semantic tracing will be introduced separately in a follow-up PR built on this branch.

## Compatibility

- The `node` feature is enabled by default.
- Existing Node consumers require no configuration changes.
- The existing Node transform interface and validation behavior are preserved.
- Rust embedders can use `--no-default-features`.

## Testing

- `cargo test --manifest-path packages/compiler/Cargo.toml`
- `cargo test --manifest-path packages/compiler/Cargo.toml --no-default-features --test host_independent_interface`
- `cargo clippy --manifest-path packages/compiler/Cargo.toml --no-default-features -- -D warnings`
- `pnpm --filter @dom-expressions/compiler build:debug`


ryansolid:
Merged with the rebase we carried (twice — `next` moved again today with the server-components behavior-claims work, so the second pass threaded the new `serverComponents` SSR option through `CompileOptions` and the adapter; codegen stayed byte-identical across all fixture/option combinations both times).

The unstable fencing stands as written: the Rust `compile()` surface carries no semver commitment pre-1.0, pin an exact revision when embedding, and `transform()` remains the supported contract. If you'd like the wording adjusted, follow-up PRs welcome.

And a standing invitation: before you build the semantic-tracing follow-up, please share the design early — especially the export contract (what gets traced, its representation, and how it stays feature-gated out of the Node build). The internals it would hook are the fastest-moving part of the codebase right now, and agreeing on that boundary first will spare you rebases like these. Looking forward to it and to solid-checker!

PR #554, opened 2026-07-26
Infer element type in applyRef function

m-canton:
This allows expressions like:

```ts
function Input(props: ComponentProps<"input">) {
  let input!: HTMLInputElement;
  
  return (
    <input ref={(el) => {
      input = el;
      if (typeof props.ref === "function") applyRef(props.ref, el);
    }} />
  );
}
```

without type errors.

### DIFF
--- a/.changeset/apply-ref-element-type.md
+++ b/.changeset/apply-ref-element-type.md
@@ -1,0 +1,5 @@
+---
+"@dom-expressions/runtime": patch
+---
+
+Make `applyRef` generic so callbacks typed for a concrete element (e.g. `HTMLInputElement`) type-check when forwarding `props.ref`.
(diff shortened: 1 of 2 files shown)


ryansolid:
Looks good — the generic + `NoInfer` fix is the right shape for the `RefCallback<HTMLInputElement>` vs `(element: Element) => void` contravariance failure under `strictFunctionTypes`.

One ask before merge: please add a patch changeset for `@dom-expressions/runtime`. Type-definition changes that affect consumers need one (see `.changeset/` / the bot note). Something like:

```md
---
"@dom-expressions/runtime": patch
---

Make `applyRef` generic so callbacks typed for a concrete element (e.g. `HTMLInputElement`) type-check when forwarding `props.ref`.
```

CI also still needs a maintainer workflow approval once that's up — happy to kick that off after the changeset lands.

ryansolid (changes requested):
Please add a patch changeset for `@dom-expressions/runtime` covering the `applyRef` typing change. The signature itself looks correct.

ryansolid:
Updated on your behalf: rebased onto current \`next\` and added a patch changeset for \`@dom-expressions/runtime\`.

ryansolid (approved):
Looks good with the changeset applied.

PR #555, opened 2026-07-27
Simplify recursive Ref type definition

m-canton:


ryansolid:
The simplification is correct — `RefCallback<T> | Ref<T>` collapses to `Ref<T>`, so `(RefCallback<T> | Ref<T>)[]` ≡ `Ref<T>[]`.

Two asks before merge:

1. **Edit `jsx-h.d.ts`, not `jsx.d.ts`.** `jsx.d.ts` is generated from `jsx-h.d.ts` via `pnpm jsx-sync-types` (see `packages/runtime/README.md`). Changes made directly to `jsx.d.ts` get discarded on the next sync. Please update the `Ref` definition in `packages/runtime/src/jsx-h.d.ts`, then run `pnpm jsx-sync-types` so both files stay in sync.

2. **Add a patch changeset** for `@dom-expressions/runtime` — type-definition changes that affect consumers need one.

ryansolid (changes requested):
Please apply the `Ref` simplification in `jsx-h.d.ts` (the source of truth) and regenerate `jsx.d.ts` with `pnpm jsx-sync-types`, plus a patch changeset for `@dom-expressions/runtime`. The type change itself is fine.

ryansolid:
Updated on your behalf:

- Applied the same `Ref` simplification in `jsx-h.d.ts` (source of truth) and confirmed `jsx.d.ts` stays in sync via `pnpm jsx-sync-types`
- Added a patch changeset for `@dom-expressions/runtime`
- Rebased onto current `next`

ryansolid (approved):
Approved — jsx-h.d.ts sync and changeset look good.

PR #558, opened 2026-08-02
Fix cached server component handoffs

brenelz:
## Summary

- restrict adopted region discovery to the owning frame namespace so nested document boundaries are not bound twice
- seed every joining frame from a rebased sibling store during cache-only handoffs
- add regressions for both failure modes

Fixes solidjs/solid#2965

## Testing

- `pnpm exec jest --no-cache --runInBand` (933 tests)
- Solid frame client suites (16 tests)
- `pnpm --dir examples/notes build`

### DIFF
--- a/.changeset/fix-cached-server-component-handoffs.md
+++ b/.changeset/fix-cached-server-component-handoffs.md
@@ -1,0 +1,5 @@
+---
+"@dom-expressions/runtime": patch
+---
+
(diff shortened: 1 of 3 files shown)


ryansolid:
Rebased onto `next` and added a runtime patch changeset. Size CI may still fail because `next`'s useHead work already exceeds size ceilings (separate from this PR).

